PR #31, opened 2026-06-15
feat(storage): add Drivine graph-backed dice-storage module for propo…

jasperblues:
### Summary
Adds a graph (Neo4j/Drivine) backend for the proposition store — ported and modernised from the
`assistant` project — and restructures `dice` into a multi-module build to house it.
`DrivinePropositionRepository` is a drop-in `PropositionRepository`, selectable against the existing
in-memory one via `embabel.dice.store.type`. Rebased on `main` after the proposition-lifecycle work
(#30), with which it integrates (decay/lifecycle).

### Build restructure
- Root becomes an aggregator `com.embabel.dice:dice-parent` (packaging `pom`).
- Existing source moves to a `dice/` module — **`com.embabel.dice:dice` coordinates are unchanged**,
  so this is not breaking for consumers (e.g. `assistant`).
- New modules: `dice-storage`, `dice-storage-autoconfigure`.

### What's in `dice-storage`
- **Graph model** (`model/`, decoupled from dice-core so KSP codegen runs with only Drivine on its
  classpath): `PropositionNode` (`@NodeFragment`, `@VectorIndex` embedding, `@RangeIndex` queryable
  fields, `@PropertyBag metadata`), `Mention`, a shared `SourceNode`, `ProcessedChunkNode`. **Two
  views over `:Proposition`** — lean `PropositionView` (mentions) for the hot paths, and
  `PropositionWithProvenanceView` (+ `DERIVED_FROM` → shared `:Source`) for `save`/`findById`.
- **`PropositionGraphMapper`** — `toView`/`toProvenanceView`/`toProposition`; enum↔name, full
  `TemporalMetadata`, `metadata` via `@PropertyBag`, provenance as shared source nodes.
- **`DrivinePropositionRepository`** — high-level `GraphObjectManager` throughout: DB-pushed
  `query(PropositionQuery)` (every filter incl. entity quantifiers), vector + entity-filtered vector,
  single-statement `findClusters`, exact-text dedup, admin methods (`reembedAll`/`clearAll`/…). One
  hand-written Cypher remains (`findByGrounding` list-membership — a flagged Drivine candidate).
- **`DrivineChunkHistoryStore`** — graph impl of dice-core's `ChunkHistoryStore`.
- **`DecayManager` / `GraphDecayManager`** — implements #30's `DecaySweeper` (lifecycle transitions),
  plus materialises `effectiveConfidence` onto nodes.
- **`dice-storage-autoconfigure`** — `embabel.dice.store.type` flip, `SchemaCatalog` (vector + range
  indexes, uniqueness constraints), and a scheduled decay tick.

### Sample: graph model (`@GraphView` + annotations)
The annotated model is the whole schema — `@VectorIndex` both declares the index *and* is what
`loadNearest` infers; `@RangeIndex` marks queryable columns; `@PropertyBag` flattens an open map to
`metadata.<key>` properties.

```kotlin
@NodeFragment(labels = ["Proposition"])
data class PropositionNode(
    @NodeId @Unique val id: String,
    @RangeIndex val contextId: String,
    val text: String,
    val confidence: Double,
    @RangeIndex val status: String,                       // PropositionStatus.name
    @RangeIndex val level: Int = 0,
    @VectorIndex(similarity = SimilarityFunction.COSINE)
    val embedding: List<Float>? = null,                   // the index loadNearest infers from
    @RangeIndex val effectiveConfidence: Double? = null,  // materialised by the decay sweep
    @PropertyBag val metadata: Map<String, Any?> = emptyMap(),   // -> metadata.<key> node properties
    // temporal flattened so TemporalMetadata round-trips:
    val validFrom: Instant? = null, val validTo: Instant? = null, val invalidatedAt: Instant? = null,
)

// shared, deduplicated source node + the per-proposition span on the edge
@NodeFragment(labels = ["Source"])
data class SourceNode(@NodeId @Unique val key: String, @RangeIndex val kind: String, /* uri/path/… */)

@RelationshipFragment
data class DerivedFrom(val chunkId: String?, val startOffset: Int?, /* … */ val source: SourceNode)

@GraphView
data class PropositionWithProvenanceView(
    @Root val proposition: PropositionNode,
    @GraphRelationship(type = "HAS_MENTION", direction = Direction.OUTGOING) val mentions: List<Mention> = emptyList(),
    @GraphRelationship(type = "DERIVED_FROM", direction = Direction.OUTGOING) val provenance: List<DerivedFrom> = emptyList(),
)
```

### Sample: repository (`GraphObjectManager`)
Filters, the entity (`HAS_MENTION`) quantifier, ordering, limit, and vector search all push into a
single Cypher statement via the generated `where { }` DSL — no whole-store scans.

```kotlin
override fun query(query: PropositionQuery): List<Proposition> =
    graphObjectManager.loadAll<PropositionView> {
        where {
            query.contextId?.let { proposition.contextId eq it.value }
            query.status?.let   { proposition.status eq it.name }
            query.entityId?.let { id -> mentions.any { resolvedId eq id } }   // relationship quantifier
        }
        orderBy { proposition.effectiveConfidence.desc() }
        query.limit?.let { limit(it) }
    }.map(PropositionGraphMapper::toProposition)
```

### Provenance: shared, queryable sources
`Proposition.provenanceEntries` persist as `(:Proposition)-[:DERIVED_FROM {chunkId, offsets, …}]->(:Source)`.
The **`:Source` node is shared** (MERGE by `SourceLocator.key()`), so a source cited by many facts is
one node — reverse-traversable ("which propositions came from this source?") and dedup'd. The
polymorphic `SourceLocator` (uri/file/content/connector) is flattened with a `kind` discriminator.
`delete` uses `DELETE_ORPHAN` so a shared source only goes when its last reference does.

### Decay
`effectiveConfidence` (time-decayed confidence) is **materialised** onto each node so confidence
ranking/filtering push into the DB. `save` seeds it (compute-on-write), and `GraphDecayManager`
refreshes it via batch write-back that recomputes through `Proposition.effectiveConfidenceAt` (single
source of truth — no decay formula re-encoded in Cypher). `DecayManager` (abstract) implements #30's
`DecaySweeper` lifecycle (ACTIVE→STALE); the autoconfigure schedules a `tick()` (materialise + sweep).

### dice-core changes
- `PropositionStoreType { IN_MEMORY, STORED }` + `storeType` on `PropositionRepository` (chat-store
  parity; default `IN_MEMORY`, non-breaking).
- `DecayManager` / `InMemoryDecayManager` (the storage-agnostic lifecycle base + no-op materialiser).
- Fixes over the legacy `assistant` store: `level`/`reinforceCount` are now actually persisted.

### Toolchain notes
- Storage modules compile with **Kotlin 2.2** (Drivine's generated DSL uses context parameters);
  dice-core stays 2.1.10 (tuProlog pin). 2.2 reads 2.1.10 metadata.
- The Drivine `where{}` DSL is generated by a nested Gradle/KSP project (`codegen-gradle/`) wired into
  Maven `generate-sources` — same pattern as `embabel-chat-store`.
- **Requires Drivine ≥ 0.0.54.**

### Status
- ✅ Full reactor builds + installs; **9 Neo4j-testcontainer integration tests pass**: save/dedup,
  full-field round-trip (incl. provenance), DB-pushed query with entity filter, vector search,
  single-Cypher clustering, chunk history, lifecycle decay sweep, effective-confidence
  materialisation/ordering, and shared-source dedup.
- ✅ All 649 dice-core tests pass post-rebase.

### Follow-ups
- More dice-storage tests: an autoconfigure context test (wires only via the hand-rolled `TestApplication`
  today) and broader `query()` filter coverage.
- Save robustness: orphan-mention cleanup on detached save; embed only on text change; `clearAll`
  leaves orphaned `:Source` nodes; `findByGrounding` list-membership as a Drivine candidate.
- `EntityMention.hints` not yet persisted.

### Integration
`dice-storage/INTEGRATE-INTO-ASSISTANT.md` walks the `assistant` migration (the real acceptance test).


dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+                query.maxLevel?.let { proposition.level lte it }
+                query.createdAfter?.let { proposition.created gte it }
+                query.createdBefore?.let { proposition.created lte it }
+                query.revisedAfter?.let { proposition.contentRevised gte it }
+                query.revisedBefore?.let { proposition.contentRevised lte it }
+                query.accessedAfter?.let { proposition.lastAccessed gte it }
+                query.accessedBefore?.let { proposition.lastAccessed lte it }
+                query.minImportance?.let { proposition.importance gte it }
+                query.minReinforceCount?.let { proposition.reinforceCount gte it }
+                query.minEffectiveConfidence?.let { proposition.effectiveConfidence gte it }
```
jimador: This applies minEffectiveConfidence against the materialized effectiveConfidence column, which is only as current as the last decay sweep, hourly by default, and was seeded at write time with the default k=2.0.

That means a query with a non-default decayK or explicit effectiveConfidenceAsOf is ignored here. The in-memory backend computes this live with effectiveConfidenceAt(asOf, decayK), so the graph and in-memory backends can return different result sets for the same input.

Can we either recompute this in Cypher when asOf or decayK are set, or make the API contract explicit that the graph backend filters/ranks against last-swept confidence only?

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+                query.entityId?.let { id -> mentions.any { resolvedId eq id } }
+                query.anyEntityIds?.let { ids -> mentions.any { resolvedId inList ids } }
+                query.allEntityIds?.forEach { id -> mentions.any { resolvedId eq id } }
+            }
+            orderBy {
+                when (query.orderBy) {
+                    OrderBy.EFFECTIVE_CONFIDENCE_DESC -> orderByEffectiveConfidenceDescNullsLast()
```
jimador: Same materialized effectiveConfidence column concern. Ordering uses the swept column, not the per-query `decayK`/`asOf`.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+                query.maxLevel?.let { proposition.level lte it }
+                query.createdAfter?.let { proposition.created gte it }
+                query.createdBefore?.let { proposition.created lte it }
+                query.revisedAfter?.let { proposition.contentRevised gte it }
+                query.revisedBefore?.let { proposition.contentRevised lte it }
+                query.accessedAfter?.let { proposition.lastAccessed gte it }
+                query.accessedBefore?.let { proposition.lastAccessed lte it }
+                query.minImportance?.let { proposition.importance gte it }
+                query.minReinforceCount?.let { proposition.reinforceCount gte it }
+                query.minEffectiveConfidence?.let { proposition.effectiveConfidence gte it }
```
jimador: Same on the vector-search path: `minEffectiveConfidence` is pushed onto the swept column rather than computed live.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+            mapOf("prefix" to contextIdPrefix),
+        )
+
+    /** Count-then-DETACH-DELETE the propositions bound by [matchClause]. Returns the number deleted. */
+    private fun deleteMatching(matchClause: String, params: Map<String, Any?>): Int {
+        val count = persistenceManager.getOne(
+            bound(QuerySpecification.withStatement("$matchClause RETURN count(p) AS count"), params)
+                .transform(Long::class.java)
+        ).toInt()
+        persistenceManager.execute(bound(QuerySpecification.withStatement("$matchClause DETACH DELETE p"), params))
```
jimador: `DETACH DELETE p` removes the proposition and its relationships, but it leaves the connected `:Mention` and `:Source` nodes behind as orphans. That bypasses the `DELETE_ORPHAN` cascade that the single-row `delete(id)` path relies on.

The integration test seems to hint at this, maybe? The `@AfterEach` has to delete `:Source` separately, and it never cleans up `:Mention`. I'm curious if this might lead to an unbounded node leak.

Can we make the bulk clear path cascade-aware too, either by deleting now-orphaned Mention/Source nodes after the proposition delete, or by routing through the same deletion path as delete(id)? Maybe also add a test that asserts `clearAll` leaves no orphaned mentions or sources behind.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+            )
+            processed++
+        }
+        logger.info("reembedAll done: propositions={} model={}", processed, embeddingService.name)
+        return processed
+    }
+
+    @Transactional
+    override fun clearAll(): Int =
+        deleteMatching("MATCH (p:Proposition)", emptyMap())
```
jimador: note: `clearAll` (and `clearByContext`/`clearByContextPrefix` below) all funnel through `deleteMatching` (see the orphan-leak note on the `DETACH DELETE` line)

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+                query.revisedAfter?.let { proposition.contentRevised gte it }
+                query.revisedBefore?.let { proposition.contentRevised lte it }
+                query.accessedAfter?.let { proposition.lastAccessed gte it }
+                query.accessedBefore?.let { proposition.lastAccessed lte it }
+                query.minImportance?.let { proposition.importance gte it }
+                query.minReinforceCount?.let { proposition.reinforceCount gte it }
+                query.minEffectiveConfidence?.let { proposition.effectiveConfidence gte it }
+                query.entityId?.let { id -> mentions.any { resolvedId eq id } }
+                query.anyEntityIds?.let { ids -> mentions.any { resolvedId inList ids } }
+                query.allEntityIds?.forEach { id -> mentions.any { resolvedId eq id } }
```
jimador: Does each `mentions.any { resolvedId eq id }` in this `forEach` compile to an independent, `AND`-ed `EXISTS` (so `allEntityIds` means "mentions all of them"), or could repeated `any` on the same relationship alias one match and collapse the constraint / make it last-write-wins?

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+     *
+     * Propositions with blank text have nothing to dedup on and are persisted directly.
+     */
+    override fun save(proposition: Proposition): Proposition {
+        val text = proposition.text
+        if (text.isBlank()) {
+            return txTemplate.execute { doPersist(proposition) }!!
+        }
+        val contextId = proposition.contextId.value
+        return synchronized(lockFor(contextId, text)) {
```
jimador, line 103: The lock + transaction is fine inside one JVM, but it does not protect the invariant across multiple instances.

Two instances pointed at the same Neo4j can both pass the existence check and insert duplicate (contextId, text) rows. `SchemaCatalog` only enforces uniqueness on id / key, so the database is not actually protecting this constraint.

What do you think about moving the invariant into Neo4j with a (contextId, text) uniqueness constraint / MERGE so it still holds under horizontal scaling?

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+        graphObjectManager.loadAll<PropositionView> {
+            where { proposition.contextId eq contextId.value }
+        }.map(PropositionGraphMapper::toProposition)
+
+    /**
+     * Grounding is a list property; "chunkId IN p.grounding" is a list-membership predicate the
+     * high-level DSL doesn't express yet. DRIVINE-CANDIDATE: list-contains in `where { }`.
+     */
+    @Transactional(readOnly = true)
+    override fun findByGrounding(chunkId: String): List<Proposition> {
```
jimador: `findByGrounding` does the lookup in hand-written Cypher, but then calls `ids.mapNotNull(::findById)`, making this a 1 + N query path.

The scan itself is already documented as the one exception to “no whole-store scans.” The avoidable issue is the N+1, especially since each `findById` loads the full `PropositionWithProvenanceView`.

A single-query fix needs to choose the contract: return lean `PropositionView` results, consistent with `findAll` / `query`, or extend the Cypher to load `DERIVED_FROM` / `:Source` and keep the current provenance behavior.

Also, if callers usually know the context, an optional `contextId` would let us bound the lookup instead of scanning every proposition.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/DiceStorageAutoConfiguration.kt
```diff
@@ -0,0 +1,196 @@
...
+    open fun graphDecayManager(
+        repository: PropositionRepository,
+        persistenceManager: PersistenceManager,
+    ): DecayManager = GraphDecayManager(repository, persistenceManager)
+
+    @Bean
+    @ConditionalOnBean(Ai::class)
+    @ConditionalOnProperty(prefix = "embabel.dice.store", name = ["type"], havingValue = "graph")
+    open fun propositionConstraintSchema(): SchemaCatalog = SchemaCatalog.of(
+        UniquenessConstraintSpec(label = "Proposition", property = "id"),
```
jimador, line 110: Uniqueness is declared on `id`/`key` only. Nothing stops a second writer from inserting a duplicate `(contextId, text)`. See the dedup note in `DrivinePropositionRepository.save`

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+                )
+                .bind(mapOf("contextId" to contextId, "text" to text, "excludeId" to excludeId))
+                .transform(String::class.java)
+        )
+
+    private fun lockFor(contextId: String, text: String): Any =
+        dedupLocks[Math.floorMod("$contextId $text".hashCode(), DEDUP_STRIPES)]
+
+    @Transactional(readOnly = true)
+    override fun findById(id: String): Proposition? =
```
jimador, line 206: `findById` returns provenance, while `query` / `findAll` return lean `PropositionView` results with empty provenance.

That looks deliberate and is documented in the `save` docstring, but only as write-cascade rationale, not as a read contract. So callers get no signal that `query(...).first()` and `findById(sameId)` can return different-looking objects.

Can we make that explicit in the type/API surface, either with distinct return types or a `provenanceLoaded` flag, so this behavior is explicit?

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -0,0 +1,433 @@
...
+    override fun count(): Int = graphObjectManager.count<PropositionView>().toInt()
+
+    /**
+     * Re-embed every proposition by writing a fresh vector onto each node. Lighter than the
+     * interface default (which re-saves the whole view): it SETs only `embedding`, leaving mentions
+     * and other properties untouched. The `@VectorIndex`-declared index is owned by Drivine, so a
+     * same-dimension re-embed needs no index DDL here.
+     */
+    @Transactional
+    override fun reembedAll(): Int {
```
jimador, line 440: This does one execute per row inside a single `@Transactional` across the whole store. For large stores, that gives us the worst of both worlds: a long-running transaction and a per-row database round trip.

Can we switch this to chunked commits or a batched `UNWIND` path so large clears/updates don’t hold one giant transaction open?

jimador (changes requested):
This looks great. Super clean, and the regression tests are rock solid.

I left a few small comments. Two are correctness issues I think we should resolve before merge; the rest are minor. Happy to approve once those are sorted.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Fixed by recomputing live when the materialized column doesn't apply. `query()` now branches on `needsLiveDecay = (decayK != 2.0 || effectiveConfidenceAsOf != null) && (minEffectiveConfidence != null || orderBy == EFFECTIVE_CONFIDENCE_DESC)`. Default case keeps the fast DB-pushed path on the swept column; otherwise it pushes every non-decay filter into Cypher and applies `effectiveConfidenceAt(asOf, decayK)` in memory over that bounded candidate set — same formula as the in-memory backend, so the two agree. New test `query with non-default decay computes effective confidence live, matching in-memory` asserts parity.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Same fix. `EFFECTIVE_CONFIDENCE_DESC` with a non-default `decayK`/`asOf` now sorts on the live `effectiveConfidenceAt(asOf, decayK)` over the candidate set rather than the swept column.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Handled. `findSimilarWithScores(request, query)` forks: when `needsLiveDecay`, it delegates to the interface default (vector search ∩ `query()`), which routes through the corrected live-decay `query()`. The fast `loadNearest { where { } }` path is kept only when decay is default.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Bulk clear is now cascade-aware: deletes the matched propositions and their `:Mention` nodes, then prunes any `:Source` left with no remaining `DERIVED_FROM` edge (shared sources referenced elsewhere survive). The Cypher moved into a standalone, parameterized resource (`queries/clear_propositions.cypher`) so it can be run/benchmarked directly. New test `clearAll removes propositions, mentions, and orphaned sources` asserts zero of each remain.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: All three funnel through one path — now `clearMatching(contextId, contextIdPrefix)`, both filters optional — backed by the single parameterized `clear_propositions.cypher`, so the orphan-aware delete covers every clear variant.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Independent AND-ed `EXISTS` — each `any{}` renders its own existential subquery, so `allEntityIds` means "mentions all of them" (not last-write-wins). Wasn't covered by a test, so added `query allEntityIds matches only propositions mentioning every id`: a prop mentioning {a,b} and one mentioning {a,b,c} match `[a,b]`; a prop mentioning only {a} does not.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: The JVM lock is only a fast path. Added a composite `(:Proposition){contextId, text}` uniqueness constraint to the `SchemaCatalog`; `save` catches the constraint violation from a concurrent cross-instance insert and reuses the existing node. The invariant holds under horizontal scaling, with the in-JVM lock still avoiding the round-trip in the common case. Caveat: Neo4j has index-key size limits for very long string properties — fine for proposition text in practice, worth keeping in mind.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/DiceStorageAutoConfiguration.kt
jasperblues: Added `UniquenessConstraintSpec("Proposition", listOf("contextId", "text"))` to the constraint `SchemaCatalog` here (and in the test `TestApplication`). Paired with catch-and-reuse in `save` — see the dedup thread.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Collapsed to a single lean query. Added a list-membership predicate to Drivine (`hasItem`, 0.0.56), so this is `loadAll<PropositionView> { where { proposition.grounding hasItem chunkId } }` — no scan, no N+1, lean `PropositionView` consistent with `findAll`/`query`. The optional `contextId`-bounded overload is a follow-up — its own small change rather than folded in here.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Made explicit on the interface. Added `withProvenance` overloads — `findAll(withProvenance)` and `query(query, withProvenance)` — defaulting to lean, plus a "provenance read contract" doc: provenance is guaranteed only on `findById` and the `withProvenance = true` overloads; plain bulk reads may omit it. The graph backend loads `PropositionWithProvenanceView` for the rich path. Went with a flag over distinct types since both map to the same `Proposition` (the in-memory backend already carries provenance, so the flag is a no-op there). New test `withProvenance loads entries only when requested`.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jasperblues: Switched to Drivine's `executeBatch` (0.0.55): embeddings are computed in memory, then written in a single batched round trip instead of one `execute` per row, eliminating the N round trips. Still one transaction — good for ~10–50K rows. I'll push a follow-up commit documenting that bound; if larger stores need it, we can add a periodic-commit variant.

jimador (approved):
Looks awesome. Great work!

PR #47, opened 2026-06-19
Store layer + trust authority

### LINKED ISSUE #12: Standalone conflict detection for propositions
## Observation

`PropositionReviser` classifies incoming propositions as `CONTRADICTORY` when they conflict with existing ones, but this is part of the revision pipeline — it requires the full LLM classification flow and produces a `RevisionResult`. There's no way to ask "does this proposition conflict with existing knowledge?" outside of revision.

Use cases that need conflict detection without full revision:

- **Pre-persistence validation** — check before committing
- **Real-time monitoring** — detect contradictions in a conversation stream without modifying the repository
- **Batch auditing** — scan existing propositions for internal contradictions
- **Domain-specific rules** — some applications need lexical/structural conflict checks (negation detection) alongside semantic ones, with different performance/cost tradeoffs

Additionally, `PropositionReviser.classify()` treats all conflicts as binary (`CONTRADICTORY` or not). Real-world conflicts have nuance — a "revision" (updating a fact) differs from a "contradiction" (asserting the opposite), which differs from "world progression" (the situation changed, not an error).

## What DICE already has

- **`PropositionReviser`** — `classify()` returns `ClassifiedProposition` with `PropositionRelation.CONTRADICTORY`, but it's coupled to the revision pipeline
- **`PropositionRelation`** enum — `IDENTICAL`, `SIMILAR`, `UNRELATED`, `CONTRADICTORY`, `GENERALIZES`
- **`RevisionResult.Contradicted`** — stores both original and new proposition
- **`PropositionRepository.findSimilar()`** — vector similarity search that could identify candidates for conflict checking

## The question

Should conflict detection be separable from revision?

Some possibilities:

1. **Extract from PropositionReviser** — make `classify()` usable standalone. It's already an interface method, but it returns `ClassifiedProposition` tied to the revision model. Could be sufficient if the only need is "check for conflicts before revising."

2. **ConflictDetector SPI** — a separate interface for conflict-only checks:
   ```kotlin
   interface ConflictDetector {
       fun detect(text: String, candidates: List<Proposition>): List<ConflictResult>
   }
   ```
   With implementations ranging from cheap (negation keyword detection + text overlap) to expensive (LLM semantic analysis). A composite detector could chain cheap-then-expensive.

3. **Richer conflict types** — extend `PropositionRelation` or introduce a `ConflictType` that distinguishes contradiction, revision, and world progression. This would benefit `PropositionReviser` too.

4. **Detection quality tracking** — when LLM-based detection fails (parse error, timeout), signal degraded quality rather than silently returning no conflicts. Consumers can route degraded results to human review.

## Open questions

- **Is separating detection from revision worth the API surface?** If most consumers run conflict detection as part of revision anyway, a standalone SPI may be overhead.
- **How reliable is negation-based detection?** Cheap lexical checks ("not", "never", "no longer") catch obvious contradictions but miss semantic ones ("the door is locked" vs. "you can walk through the door").
- **Should conflict resolution be part of this?** Detection tells you *that* a conflict exists. Resolution tells you *what to do about it*. These could be separate SPIs or combined.

### LINKED ISSUE #13: Proposition priority and authority model
## Observation

All propositions in `PropositionRepository` are effectively peers. `confidence` and `importance` exist, but there's no framework-level concept of which propositions matter most — which should be retained vs. evicted under pressure, which should win in conflicts, or which an LLM should comply with more strongly.

This matters when:

- **Token budgets are limited** — not all propositions fit in a prompt; which get included?
- **Knowledge conflicts arise** — two conflicting propositions exist; which wins?
- **Memory pressure** — too many propositions; which get evicted?
- **Compliance directives** — some facts must be respected ("patient is allergic to penicillin") while others are suggestions

## What DICE already has

- **`confidence: ZeroToOne`** — LLM's extraction-time certainty. Doesn't evolve post-extraction (except through `effectiveConfidence()` decay).
- **`importance: ZeroToOne`** — LLM's assessment of how much a fact matters. Set at extraction time, doesn't change.
- **`decay: ZeroToOne`** — staleness rate. Feeds `effectiveConfidence()` but doesn't drive behavior beyond that.
- **`reinforceCount: Int`** — tracks merge/reinforcement frequency. Stored but not acted on.
- **`status: PropositionStatus`** — ACTIVE, SUPERSEDED, CONTRADICTED, PROMOTED. Lifecycle state, not priority.

The gap: `confidence` and `importance` are extraction-time snapshots. There's no lifecycle property that evolves based on reinforcement, decay, conflict resolution, or operator action.

## The question

Is `confidence` + `importance` sufficient, or do propositions need a priority model that evolves over their lifecycle?

Some possibilities:

1. **Make existing fields dynamic** — instead of new fields, let `confidence` and `importance` be updated by reinforcement, decay, and revision. `reinforceCount` already tracks frequency; a policy could boost `confidence` based on it.

2. **Add a `rank` field** — an integer priority score (e.g., 0 = unranked, higher = more important). Influences retrieval ordering, eviction priority, conflict resolution. Evolves over time based on reinforcement and decay.

3. **Add an `authority` level** — an enum indicating compliance strength, e.g.:

   | Level | Semantics |
   |-------|-----------|
   | PROVISIONAL | May consider — unverified, low confidence |
   | RELIABLE | Should respect — well-established |
   | CANON | Must respect — ground truth, never auto-demoted |

   Authority transitions: reinforcement promotes (PROVISIONAL → RELIABLE), decay demotes (RELIABLE → PROVISIONAL), CANON requires explicit operator action.

4. **Rank + authority together** — rank = relative importance within an authority tier. Authority = how strongly the LLM should comply. Orthogonal dimensions.

### Why not just use `importance`?

`importance` is a 0-1 float set at extraction time reflecting the LLM's assessment. Rank and authority would be lifecycle properties that evolve based on what happens *after* extraction:

- `importance`: "How important did the extractor think this was?"
- `rank`: "How important is this proposition now, given everything since extraction?"
- `authority`: "How strongly should an LLM comply with this?"

### LINKED ISSUE #34: A pluggable proposition store that survives a restart
The proposition store is one monolithic `PropositionRepository` — every backend has to implement the
whole surface, even capabilities it can't actually support, and the only backend is in-memory so
nothing survives a restart.

Split it into small capability fragments behind a `PropositionStore` facade, so a backend declares
only what it honestly supports — the way embabel-agent splits its RAG `SearchOperations` into opt-in
`VectorSearch` / `TextSearch` / `CoreSearchOperations`, and types its conversation store with
`ConversationStoreType` (IN_MEMORY / STORED). Then add a durable file-backed repository alongside
the in-memory one.

## Use cases

- Swap the in-memory store for the durable file-backed one so knowledge survives a restart, with no
  pipeline changes.
- Bring up a new backend that only does plain lookups without being forced to fake the search or
  ranking it can't honestly support.
- Run fast tests against the in-memory fragment while production uses the durable repository.

### LINKED ISSUE #35: Know which facts to trust when they disagree
Propositions carry no sense of how much to trust them. We don't know whether a fact came from an
authoritative source or a derived guess, and when two propositions clash there's nothing that
notices or decides which one wins — the reviser just merges blindly.

Add source-authority tiers, an advisory trust score, and conflict detection, so the reviser and the
downstream gates, consolidation, and projection can prefer authoritative, non-conflicting facts
instead of treating everything as equally true. The score is advisory — it informs decisions rather
than deleting anything, so nothing is lost, it's just ranked.

## Use cases

- A fact pulled from an official policy document outweighs one inferred from a passing chat mention
  when the two disagree.
- A reviewer can see, before relying on an answer, whether its supporting facts came from
  authoritative sources or low-confidence guesses.
- Consolidation and projection skip or down-weight conflicting, low-trust propositions instead of
  baking them into the graph as settled truth.

jimador:
## Store layer + trust authority

First of a four-PR stack that lands the knowledge-graph feature set on the restructured multi-module
layout.

### What's in it

**A pluggable proposition store that survives a restart** (#34)
- Splits the monolithic `PropositionRepository` into small capability fragments behind a
  `PropositionStore` facade, so a backend declares only what it honestly supports — mirroring how
  embabel-agent splits RAG `SearchOperations` into opt-in `VectorSearch` / `TextSearch` /
  `CoreSearchOperations` and types its conversation store with `ConversationStoreType`.
- Adds a durable file-backed repository alongside the in-memory one, so knowledge survives a restart.

**Know which facts to trust when they disagree** (#35)
- Source-authority tiers (`Authority`), an advisory trust score (`TrustScorer` +
  `AuthorityWeightedTrustScorer`), and standalone conflict detection (`ConflictDetector` /
  `ConflictType`), so the reviser and the downstream gates, consolidation, and projection can prefer
  authoritative, non-conflicting facts instead of treating everything as equally true.
- The trust score is advisory — it informs decisions rather than deleting anything.

### Resolves pre-existing issues

The trust-authority work answers three open design questions directly:

- **#12 Standalone conflict detection** — `ConflictDetector` makes conflict checking usable outside
  the revision pipeline, and `ConflictType` distinguishes contradiction from revision/progression.
- **#13 Proposition priority and authority model** — `Authority` adds the compliance-strength tier
  the issue proposed, and conflict resolution can prefer higher-authority propositions.
- **#14 Post-extraction trust scoring** — `TrustScorer` scores propositions on signals beyond
  extraction confidence; `AuthorityWeightedTrustScorer` is a configurable built-in. (The gate that
  wires this into the pipeline — #15 — lands in the next PR.)

### Notes

- The Docker `api.version=1.41` Testcontainers pin lives in the parent pom (added here), since the
  inherited `dice-storage` module's Drivine integration test needs it — so plain `mvn verify` runs
  all ITs with no flags.
- Builds on `knowledge-events` and `proposition-lifecycle`, already on `main`.

Closes #34
Closes #35
Closes #12
Closes #13
Closes #14


dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    /** Indirect or derived source (e.g. content-addressed material, abstractions). */
+    DERIVED,
+
+    /** Authority could not be determined; the lowest-authority fail-safe default. */
+    UNKNOWN,
+}
+
+/**
+ * Policy SPI: resolves the [AuthorityTier] of a proposition's source.
```
igordayen, line 50: maybe consider SPI sub-package

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
```diff
@@ -0,0 +1,120 @@
...
+     * @param proposition The proposition whose source authority to resolve
+     * @return the resolved [AuthorityTier]
+     */
+    fun resolve(proposition: Proposition): AuthorityTier
+}
+
+/**
+ * [AuthorityResolver] that always returns a single configured [tier], ignoring the
+ * proposition entirely. Useful when every proposition in a context shares the same
+ * provenance, or as a deterministic stand-in.
```
igordayen, line 70: is semantic "fall-back"?

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @property tierByLocatorKind Mapping from [SourceLocator] subtype to [AuthorityTier];
+ *   defaults to [DEFAULT_MAP]
+ */
+class StructuralAuthorityResolver @JvmOverloads constructor(
+    private val tierByLocatorKind: Map<KClass<out SourceLocator>, AuthorityTier> = DEFAULT_MAP,
+) : AuthorityResolver {
+
+    override fun resolve(proposition: Proposition): AuthorityTier {
+        val locators = proposition.provenanceEntries.map { it.locator }
```
igordayen, line 99: logging for output could be helpful for debugging

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    companion object {
+        /**
+         * Default locator-kind → tier mapping: connector references are first-party
+         * ([AuthorityTier.PRIMARY]); URIs and files are named external sources
+         * ([AuthorityTier.SECONDARY]); content-addressed material is indirect
+         * ([AuthorityTier.DERIVED]).
+         */
+        @JvmField
+        val DEFAULT_MAP: Map<KClass<out SourceLocator>, AuthorityTier> = mapOf(
```
igordayen, line 124: perhaps consider externalization for future evolution

dice/src/main/kotlin/com/embabel/dice/spi/AuthorityWeightedTrustScorer.kt
```diff
@@ -0,0 +1,66 @@
...
+    companion object {
+
+        /** Neutral half-confidence used when the source authority is unknown. */
+        const val DEFAULT_UNKNOWN_SCORE: Double = 0.5
+
+        /**
+         * Sensible defaults: trust falls as the source gets more indirect, and an unknown source
+         * lands at the neutral midpoint.
+         */
+        val DEFAULT_WEIGHTS: Map<AuthorityTier, Double> = mapOf(
```
igordayen, line 63: please consider configurability

dice/src/main/kotlin/com/embabel/dice/spi/TrustScorer.kt
```diff
@@ -0,0 +1,104 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.common
+
+import com.embabel.dice.proposition.Proposition
+import com.embabel.dice.proposition.revision.ConflictType
+
+/**
+ * Policy SPI: assigns a trust score to a proposition.
```
igordayen, line 21: SPI sub-package?

dice/src/main/kotlin/com/embabel/dice/spi/ConflictDetector.kt
```diff
@@ -0,0 +1,127 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.proposition.revision
+
+import com.embabel.dice.proposition.Proposition
+import java.time.Instant
+
+/**
+ * Policy SPI: classifies the [ConflictType] between an incoming proposition and an
```
igordayen, line 23: same

dice/src/main/kotlin/com/embabel/dice/spi/ConflictDetector.kt
```diff
@@ -0,0 +1,127 @@
...
+ * @property evolvingPredicates Lower-cased predicate names treated as time-evolving;
+ *   defaults to [DEFAULT_EVOLVING_PREDICATES]
+ */
+class TemporalConflictDetector @JvmOverloads constructor(
+    private val evolvingPredicates: Set<String> = DEFAULT_EVOLVING_PREDICATES,
+) : ConflictDetector {
+
+    override fun detect(incoming: Proposition, existing: Proposition): ConflictType {
+        val predicate = (
+            (incoming.metadata[Proposition.PREDICATE] as? String)
```
igordayen, line 98: safe? What are potentially other types

dice/src/main/kotlin/com/embabel/dice/proposition/revision/ConflictDetector.kt
```diff
@@ -0,0 +1,127 @@
...
+    override fun detect(incoming: Proposition, existing: Proposition): ConflictType {
+        val predicate = (
+            (incoming.metadata[Proposition.PREDICATE] as? String)
+                ?: (existing.metadata[Proposition.PREDICATE] as? String)
+            )?.lowercase()
+            ?: return ConflictType.Contradiction
+        if (predicate !in evolvingPredicates) return ConflictType.Contradiction
+        // Equal timestamps are not a temporal contradiction (neither strictly supersedes),
+        // so only a strictly-older incoming falls back to Contradiction.
+        return if (recencyOf(incoming).isBefore(recencyOf(existing))) {
```
igordayen: any logging

dice/src/main/kotlin/com/embabel/dice/spi/ConflictDetector.kt
```diff
@@ -0,0 +1,127 @@
...
+         * Default set of predicate names treated as time-evolving facts. Kept small
+         * and configurable; consumers can supply their own set.
+         *
+         * Note: `"status"` here is a free-text *predicate* string (a fact like "status: married"),
+         * unrelated to the [com.embabel.dice.proposition.PropositionStatus] enum — the name overlap is
+         * coincidental.
+         */
+        @JvmField
+        val DEFAULT_EVOLVING_PREDICATES: Set<String> =
+            setOf("employer", "residence", "status", "role", "location", "title")
```
igordayen, line 137: hardcoded for tests?

dice/src/main/kotlin/com/embabel/dice/proposition/revision/LlmPropositionReviser.kt
```diff
@@ -236,15 +289,15 @@ data class LlmPropositionReviser(
...
                 contextId = newProposition.contextId,
-                status = PropositionStatus.ACTIVE,
+                statuses = setOf(PropositionStatus.ACTIVE),
             )
         )
         val canonicalMatch = contextProps.find { canonicalize(it.text) == newCanonical }
         if (canonicalMatch != null) {
             val original = repository.findById(canonicalMatch.id) ?: canonicalMatch
             val merged = mergePropositions(original, newProposition)
             logger.debug("Canonical text match: {}", newProposition.text.take(60))
```
igordayen, line 302: excessive logging?

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @param path the JSON file the store is persisted to and loaded from.
+ * @param embeddingService optional embedder; when null, vector paths return empty results.
+ */
+class JsonFilePropositionRepository @JvmOverloads constructor(
+    private val path: Path,
+    private val embeddingService: EmbeddingService? = null,
+) : PropositionRepository {
+
+    private val mapper = jacksonObjectMapper().findAndRegisterModules()
```
igordayen, line 58: FYI- jackson is significantly upgraded fo boot 4.x/sping ai 2.x - package structure changes

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
```diff
@@ -0,0 +1,235 @@
...
+
+    private val propositions = ConcurrentHashMap<String, Proposition>()
+    private val embeddings = ConcurrentHashMap<String, FloatArray>()
+
+    // Serializes the mutate+flush critical section so concurrent writers cannot persist a
+    // torn/stale snapshot and so in-memory state can be rolled back if the disk flush fails.
+    private val writeLock = Any()
+
+    init {
+        if (Files.exists(path)) {
```
igordayen, line 68: I assume the file will not survive container re-deployment in a cloud environment.  you may need NAS?

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
```diff
@@ -0,0 +1,235 @@
...
+            .mapNotNull { prop ->
+                val propEmbedding = embeddings[prop.id] ?: return@mapNotNull null
+                val similarity = cosineSimilarity(queryEmbedding, propEmbedding)
+                if (similarity >= minSimilarity) SimilarityResult(match = prop, score = similarity) else null
+            }
+            .sortedByDescending { it.score }
+            .take(textSimilaritySearchRequest.topK)
+    }
+
+    private fun cosineSimilarity(a: FloatArray, b: FloatArray): Double {
```
igordayen, line 163: The org.apache.commons.text.similarity.CosineSimilarity class in the [Apache Commons Text](https://commons.apache.org/proper/commons-text/javadocs/api-release/allclasses-index.html)

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
```diff
@@ -0,0 +1,235 @@
...
+    override fun count(): Int = propositions.size
+
+    /**
+     * Snapshot the full store to disk crash-safely: write to a temp file in the target's parent
+     * directory, then atomically rename it over the target. The parent directory is created if
+     * absent.
+     */
+    private fun flush() {
+        val parent = path.toAbsolutePath().parent
+        Files.createDirectories(parent)
```
igordayen, line 214: you may not have permissions in cloud env

dice/src/main/kotlin/com/embabel/dice/proposition/PropositionStore.kt
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * The gate is deliberately fail-open: a null [threshold] disables it entirely; a proposition
+ * with no cached score passes (unscored ≠ low-trust); a non-finite cached value (e.g. `NaN`)
+ * also passes rather than being silently dropped.
+ *
+ * @param threshold minimum acceptable trust score, or `null` to skip the check
+ * @return `true` if the proposition passes the trust gate
+ */
+internal fun Proposition.passesMinTrust(threshold: Double?): Boolean {
+    if (threshold == null) return true
```
igordayen, line 39: please consider idiomatic kotlin ("when")

dice/src/main/kotlin/com/embabel/dice/proposition/PropositionStore.kt
```diff
@@ -0,0 +1,218 @@
...
+ * @param prop the proposition to test
+ * @param asOf the instant at which effective confidence is evaluated; pass one value for the whole
+ *   query so filtering and ordering agree. Defaults to the query's `effectiveConfidenceAsOf`, or now.
+ * @return `true` if the proposition satisfies every active filter
+ */
+internal fun PropositionQuery.matchesFilters(
+    prop: Proposition,
+    asOf: Instant = effectiveConfidenceAsOf ?: Instant.now(),
+): Boolean {
+    if (contextId != null && prop.contextId != contextId) return false
```
igordayen, line 59: idiomatic kotlin

dice/src/main/kotlin/com/embabel/dice/proposition/VectorSearchCapable.kt
```diff
@@ -0,0 +1,120 @@
...
+     * Each cluster has an anchor proposition and a list of similar propositions
+     * above the similarity threshold. Clusters are deduplicated so that each
+     * pair appears only once (the proposition with the lower ID is the anchor).
+     *
+     * @param similarityThreshold Minimum cosine similarity to include in a cluster
+     * @param topK Maximum number of similar items per cluster
+     * @param query Optional query to pre-filter which propositions participate
+     * @return Clusters ordered by size (largest first), excluding empty clusters
+     */
+    fun findClusters(
```
igordayen, line 102: logging

dice/src/main/kotlin/com/embabel/dice/proposition/VectorSearchCapable.kt
```diff
@@ -0,0 +1,120 @@
...
+        similarityThreshold: ZeroToOne = 0.7,
```
igordayen, line 103: config

dice/src/test/kotlin/com/embabel/dice/spi/ConflictTypeTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+
+class ConflictTypeTest {
+
+    @Test
+    fun `custom accepts a non-blank label`() {
+        val custom = ConflictType.Custom("supersession")
```
igordayen, line 26: too trivial

dice/src/test/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepositoryTest.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+package com.embabel.dice.proposition.store
+
+import com.embabel.agent.core.ContextId
+import com.embabel.common.core.types.TextSimilaritySearchRequest
+import com.embabel.dice.proposition.Proposition
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
```
igordayen, line 24: im not sure whether it will work in github CI

pom.xml
```diff
@@ -91,4 +91,23 @@
...
+                <!--
+                    Pin the Docker Engine API version the Testcontainers docker-java client uses;
+                    its default is below the engine minimum and disables container-backed tests.
+                    Honoured only as a JVM system property, and tests fork, so it must be set here.
+                -->
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables>
+                        <api.version>1.41</api.version>
```
igordayen: please use meta-property

igordayen (commented):
excellent work, few comments to consider, thanks

pom.xml
jimador: good call, pulled it up into a named property in `<properties>` and referencing it from there now

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
jimador: this one's by design - it's the local/single-process reference backend, not the production durability story, that's the graph store's job. made it explicit in the class kdoc now. in a container or cloud deploy you'd wire up the Drivine/Neo4j store, or mount durable storage if you really do want the file backend

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
jimador: this one's by design - it's the local/single-process reference backend, not the production durability story, that's the graph store's job. made it explicit in the class kdoc now. in a container or cloud deploy you'd wire up the Drivine/Neo4j store, or mount durable storage if you really do want the file backend

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
jimador: good idea - added a debug line here that logs the tier it resolved plus the locator kinds it saw, so the decision is visible once you turn debug on

dice/src/main/kotlin/com/embabel/dice/proposition/revision/ConflictDetector.kt
jimador: added a debug line for this one too - it logs the predicate and which way recency went (world-progression vs contradiction), off by default but there when you need it

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
igordayen: thanks, would it pass unit test in CI on GitHub?

pom.xml
igordayen: could you please mark it as "resolved"

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
jimador: yep, it already does. the tests use junit's `@TempDir`, so they get a throwaway temp dir on the runner and clean themselves up

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
jimador: Commons Text `CosineSimilarity` works on `Map<CharSequence,Integer>` term-frequency vectors (lexical text similarity). Here we compare dense float embedding vectors.

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
jimador: I'll probably leave this for a followup PR. The mapping is already injectable via the `tierByLocatorKind` constructor arg. These just define the "defaults"

dice/src/main/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepository.kt
jimador: thanks for the heads up. this is the reference backend so it just tracks the project's boot/spring-ai version - we'll pick up the jackson package moves when the framework bump lands instead of getting ahead of it here

dice/src/test/kotlin/com/embabel/dice/proposition/store/JsonFilePropositionRepositoryTest.kt
jimador: it uses junit's `@TempDir` so it's CI-safe, and the `dice` unit tests (this one included) already pass on CI

dice/src/main/kotlin/com/embabel/dice/spi/AuthorityWeightedTrustScorer.kt
jimador: it already is - `weights` and `unknownScore` are both constructor params, `DEFAULT_WEIGHTS` is just the default. pass your own map in to retune it

dice/src/main/kotlin/com/embabel/dice/proposition/VectorSearchCapable.kt
jimador: `similarityThreshold` and `topK` are caller-overridable defaults on the method, 0.7 is just the default value

dice/src/main/kotlin/com/embabel/dice/spi/ConflictDetector.kt
jimador: not test-only - it's the default for the `evolvingPredicates` constructor arg, consumers can pass their own set in

dice/src/main/kotlin/com/embabel/dice/spi/ConflictDetector.kt
jimador: `as?` is a safe cast - a missing or non-string predicate just falls through to the existing proposition's predicate, and then to a conservative `Contradiction`. never throws

dice/src/test/kotlin/com/embabel/dice/spi/ConflictTypeTest.kt
jimador: these pin `ConflictType.Custom`'s blank/empty-label rejection which is a real constructor invariant, so they're guarding behavior rather than restating it. i'd keep them

dice/src/main/kotlin/com/embabel/dice/proposition/revision/LlmPropositionReviser.kt
jimador: that one's `logger.debug` with truncated text so it's off by default, only shows when you enable debug. comfortable leaving it

dice/src/main/kotlin/com/embabel/dice/spi/Authority.kt
jimador: yeah - `UNKNOWN` is the deliberate fail-safe fallback (lowest authority) when a tier can't be determined. happy to spell it out as 'fallback' in the wording if that's clearer

dice/src/main/kotlin/com/embabel/dice/proposition/VectorSearchCapable.kt
jimador: added some debug logging here in this PR - `findClusters` and the filtered `findSimilarWithScores` log candidate and hit counts now

dice/src/main/kotlin/com/embabel/dice/proposition/PropositionStore.kt
jimador: these are intentional early-return guard chains, kept flat and uniform so each filter's easy to scan and add to. a subjectless `when` doesn't really simplify the per-field null checks so i'd lean toward keeping it

dice/src/main/kotlin/com/embabel/dice/proposition/PropositionStore.kt
jimador: same as the other one - the guard chain keeps each filter independent and easy to extend. happy to revisit if you feel strongly though

igordayen (approved):
looks good, excellent work.

PR #48, opened 2026-06-19
Extraction concurrency + maintenance

### LINKED ISSUE #9: Proposition pinning and eviction immunity
## Observation

Under memory pressure (too many propositions for the token budget), some propositions must be retained regardless of age, confidence, or access frequency. There's no mechanism for marking a proposition as "must retain."

- "Patient is allergic to penicillin" — must never be evicted, even if old and unreferenced
- Operator-designated ground truth — facts that define the rules of the context
- Domain invariants — operating constraints the system must always respect

Without pinning, an adversarial user could indirectly evict critical facts by flooding the context with noise propositions that push important ones out of the budget.

## What DICE already has

- **`PropositionStatus`** — ACTIVE, SUPERSEDED, CONTRADICTED, PROMOTED — lifecycle state, but nothing means "never evict"
- **`importance: ZeroToOne`** — high importance doesn't guarantee retention under pressure
- **`metadata: Map<String, Any>`** — could carry `pinned: true` as a convention

## Proposal

Something like:

```kotlin
data class Proposition(
    // ... existing fields ...
    val pinned: Boolean = false,
)
```

Behavior:
- **Eviction immune** — never removed by budget enforcement or memory pressure
- **Decay immune** — `effectiveConfidence()` or decay policies (#11) skip pinned propositions
- **Always injected** — token budget enforcement (#07) includes pinned propositions first, then fills remaining budget from unpinned
- **Conflict protection** — contradicting a pinned proposition could require explicit confirmation rather than automatic resolution

Repository additions:
```kotlin
fun pin(propositionId: String): Proposition
fun unpin(propositionId: String): Proposition
fun findPinned(contextId: ContextId): List<Proposition>
```

Backward-compatible: `pinned = false` is the default.

## Open questions

- **Is a boolean sufficient?** Or do you need pinning *reasons* and *scopes* (e.g., pinned for a specific conversation vs. globally)?
- **What if pinned propositions exceed the token budget?** Should there be a hard limit on pinned propositions per context?
- **Should pinning be automatic?** E.g., operator-sourced propositions are auto-pinned, user-sourced are not.

### LINKED ISSUE #36: Extract many documents at once without breaking reconciliation
Extraction runs serially, one artifact at a time, blocking until done — a bottleneck for any
document set of real size, and the scheduling is hard-wired into the pipeline.

Separate extraction into two phases — a parallel extract phase and a serialized reconcile phase —
behind a pluggable execution strategy, so concurrency becomes a swappable policy rather than
something baked in.

## Use cases

- Ingest a large document set and have extraction fan out across artifacts instead of crawling one
  at a time.
- Keep reconciliation correct under parallelism by serializing just the merge step, not the whole
  pipeline.
- Switch policy — serial while debugging, parallel for throughput — without touching pipeline code.

### LINKED ISSUE #37: Reclaim stale and duplicate facts instead of growing forever
Stale, superseded, and duplicate propositions pile up over time with no way to reclaim them — the
store only ever grows.

Add a GC-style mark-and-sweep collector: pluggable strategies mark what looks like garbage (decayed,
duplicate), a sweep policy decides what actually happens to it (archive, delete, demote), and every
reclamation is recorded so the sweep is auditable after the fact.

## Use cases

- Reclaim the space taken by superseded and duplicate facts after a long stretch of ingestion.
- Archive rather than delete decayed propositions when you need to keep an audit trail.
- Tune what counts as garbage, and what happens to it, per deployment without rewriting the
  collector.

### LINKED ISSUE #38: Tidy and consolidate what DICE knows between sessions
Between sessions, nothing tidies up what DICE knows — related propositions stay scattered,
contradictions persist, and decayed entries accumulate. The graph only grows raw facts and never
consolidates them.

Add a background "dream loop": composable consolidation passes (abstract related propositions,
resolve contradictions, sweep decayed ones) and an orchestrator that runs them as repeatable cycles,
wired into the existing memory maintenance so it happens as normal upkeep.

## Use cases

- Overnight, abstract a cluster of related facts into a higher-level proposition so the graph gets
  more concise instead of just larger.
- Periodically resolve contradictions that piled up during the day's ingestion.
- Run consolidation on a schedule as routine upkeep — the way memory settles between sessions —
  rather than as a manual chore.

### LINKED ISSUE #39: Check new facts before they reach the store
Freshly extracted propositions reach the store unchecked — there's no checkpoint to drop junk, hold
low-confidence items for review, or keep under-evidenced relations out of the graph. Weak facts get
admitted and have to be cleaned up later.

Add a gate pipeline that runs right after extraction and emits an observable decision for each
proposition — persist, route to review, reject, skip projection, or demote a relation that falls
below the evidence floor.

## Use cases

- Drop obviously junk extractions before they ever reach the store.
- Hold low-confidence propositions in a review queue instead of admitting or discarding them
  outright.
- Keep an under-evidenced relation out of the graph by demoting it, while still retaining the
  underlying proposition.

jimador:
## Extraction concurrency + maintenance

Second of the four-PR stack. **#47 (store layer + trust authority) has merged and this branch is rebased onto `main`** — the diff below is this PR's own changes only. #49 and #50 stack on top of this one.

### What's in it

- **Extract many documents at once without breaking reconciliation** (#36) — splits extraction into
  a concurrent extract stage and a serial reconcile stage behind a pluggable execution strategy,
  so concurrency is a swappable policy rather than baked in.
- **Reclaim stale and duplicate facts instead of growing forever** (#37) — a GC-style mark-and-sweep
  collector: pluggable mark strategies, a sweep policy (transition / delete / skip), and an audit
  record for every reclamation.
- **Tidy and consolidate what DICE knows between sessions** (#38) — a background "dream loop" of
  composable consolidation passes (abstract related, resolve contradictions, sweep decayed) and an
  orchestrator that runs them as repeatable, threshold-gated cycles, wired into existing memory
  maintenance.
- **Check new facts before they reach the store** (#39) — an `ExtractionGatePipeline` that runs
  right after extraction and emits an observable decision per proposition (persist, route to review,
  reject, skip projection, or demote below the evidence floor).

### Resolves pre-existing design issues

- **#15 Post-extraction evaluation pipeline** — `ExtractionGatePipeline` / `StandardGates`
  implements the configurable gate chain the issue proposed (pass / reject / route / transform,
  cheap gates first), sitting between extraction and persistence.
- **#9 Proposition pinning and eviction immunity** — `Proposition.pinned` is now a real "must
  retain" guarantee: pinned propositions are immune to decay marking, sweep reclamation, and
  automatic contradiction (at both ingest and consolidation), with `pin`/`unpin`/`findPinned`
  store helpers and a `pinned` query filter. Budget-priority injection stays with #7.

Related but not closed here: #10 (decay sweeping advances it; content-aware decay curves remain
open).

### Hardening & quality

After the feature work the branch went through an adversarial review pass and follow-up hardening:

- **Correctness** — the REST extract endpoints now persist revised/contradicted originals (not just
  the new propositions); contradiction resolution resolves each pair once with a corrected tie-break;
  the concurrent extraction strategies are `AutoCloseable` so a self-created pool can't leak; gates
  expose a `shouldSave` guard so Demote/SkipProjection aren't dropped; dream-loop cycle totals count
  new vs. transitioned correctly; dry-run sweeps record `MARKED` rather than looking like real sweeps.
- **Concurrency** — the dream loop locks cycles per `contextId` (no overlapping same-context cycles;
  different contexts still run in parallel), the pipeline's per-chunk failure map is a
  `ConcurrentHashMap`, and the thread-safety contracts across the collector, maintenance, and
  resolver are documented.
- **Observability** — debug/trace logging across extraction, maintenance, the consolidation passes,
  and per sweep decision.
- **API placement** — the sweep-policy cluster (`SweepPolicy`/`StatusTransitionSweepPolicy` plus its
  `SweepAction`/`PropositionMark`/`MarkReason` vocabulary) lives in `com.embabel.dice.spi` alongside
  the other lifecycle/trust policies. _(Moving these public types is source-incompatible for any
  external importer of the old package — acceptable pre-1.0.)_
- **Docs** — four design notes under `docs/design/` with mermaid diagrams (extraction pipeline;
  knowledge hygiene; consolidation & the dream loop; reclamation & the collector), and AGENTS
  navigation updated for every new subsystem.
- **Tests** — per-context-lock concurrency regression tests, a `DecayStatusPolicy` unit test, and a
  no-mocks end-to-end reclamation test, on top of the per-feature suites.

### Verification

`mvn verify` is green across all modules: dice 935 unit tests + dice-storage 27 (including the Neo4j
Testcontainers integration tests).

Closes #36
Closes #37
Closes #38
Closes #39
Closes #15
Closes #9


dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
```diff
@@ -0,0 +1,203 @@
...
+        copy(changeVolumeThreshold = threshold)
+
+    override fun consolidate(contextId: ContextId): DreamLoopReport? = synchronized(lockFor(contextId)) {
+        // Hold the per-context lock across the gate check and the cycle so the threshold decision
+        // and the count update can't race a concurrent trigger for the same context.
+        val currentActive = activeSnapshot(contextId).size
+        val delta = currentActive - (lastActiveCount[contextId] ?: 0)
+        if (delta < changeVolumeThreshold) {
+            // Not enough has changed to be worth a cycle; record the count and skip.
+            lastActiveCount[contextId] = currentActive
```
jasperblues: **Scheduled consolidation never fires under normal polling.** The skip branch advances the baseline (`lastActiveCount[contextId] = currentActive`), so the gate measures delta *since the last call*, not *since the last cycle* as the KDoc says. When growth arrives in increments smaller than `changeVolumeThreshold` (the normal scheduled case), the delta never accumulates and a cycle never runs. Fix: only advance the baseline after an actual cycle (step 5 already does), leave it untouched on skip.

Failing test:
```kotlin
@Test
fun `consolidate accumulates growth across skipped calls until the threshold is reached`() {
    val repository = mockk<PropositionRepository>(relaxed = true)
    var activeCount = 0
    every { repository.query(any()) } answers { List(activeCount) { prop("p$it") } }
    val pass = FakePass("noop") { ConsolidationPassResult.NoOp("noop") }
    val orchestrator = DefaultDreamLoopOrchestrator.withRepository(repository).withPass(pass)
    repeat(5) { activeCount += 3; orchestrator.consolidate(contextId) }   // 3,6,9,12,15
    assertTrue(pass.runCount >= 1) {
        "consolidation never fired despite cumulative growth of $activeCount (threshold 10)"
    }
}
```

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
```diff
@@ -0,0 +1,203 @@
...
+        val snapshot = activeSnapshot(contextId)
+        logger.debug("Running {} pass(es) over {} active proposition(s) for {}", passes.size, snapshot.size, contextId)
+
+        // (2) Run each pass in registration order, isolating failures into Failed results.
+        val passResults = passes.map { pass -> runPass(pass, contextId, snapshot) }
+
+        // (3) Aggregate saves into a single write.
+        val snapshotIds = snapshot.mapTo(HashSet()) { it.id }
+        val changed = passResults.filterIsInstance<ConsolidationPassResult.Changed>()
+        val toSave: List<Proposition> = changed.flatMap { it.propositionsToSave }
```
jasperblues: **No cross-pass reconciliation → nondeterministic final status.** All passes run over the same snapshot and their saves are flat-mapped with no dedup. The same proposition returned by `AbstractionPass` (`SUPERSEDED`) and `ContradictionResolutionPass` (`CONTRADICTED`) in one cycle is written twice; the persisted result depends on `saveAll` ordering. Suggest a `distinctBy { it.id }` with a defined precedence before the write.

Failing test:
```kotlin
@Test
fun `aggregated saves reconcile to a single status per proposition id`() {
    val repository = mockk<PropositionRepository>(relaxed = true)
    val original = prop("shared")
    every { repository.query(any()) } returns listOf(original)
    val captured = slot<Collection<Proposition>>()
    every { repository.saveAll(capture(captured)) } returns Unit
    DefaultDreamLoopOrchestrator.withRepository(repository)
        .withPass(FakePass("abstraction") { ConsolidationPassResult.Changed("abstraction",
            propositionsToSave = listOf(original.withStatus(PropositionStatus.SUPERSEDED))) })
        .withPass(FakePass("contradiction") { ConsolidationPassResult.Changed("contradiction",
            propositionsToSave = listOf(original.withStatus(PropositionStatus.CONTRADICTED))) })
        .consolidateNow(contextId)
    assertEquals(1, captured.captured.count { it.id == original.id })
}
```

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
```diff
@@ -0,0 +1,203 @@
...
+        lastActiveCount[contextId] = activeSnapshot(contextId).size
+
+        // (6) Build the report.
+        // A "new" proposition is one whose id was not in the examined snapshot (e.g. a fresh
+        // abstraction); a re-saved snapshot member (e.g. a source marked SUPERSEDED, or a
+        // CONTRADICTED loser) is a transition, not a new proposition. Decay transitions are written
+        // by their own pass, so they are counted via externallyApplied rather than the save list.
+        val newPropositions = toSave.count { it.id !in snapshotIds }
+        val savedTransitions = toSave.count { it.id in snapshotIds }
+        val transitioned = savedTransitions +
```
jasperblues: **Deletes are counted as transitions even when never applied.** `totalTransitioned` adds `propositionsToDelete.size` unconditionally, but deletes only execute when `allowHardDelete` is true (default false). The report over-counts transitions that didn't happen. Latent for this PR's passes (all return empty delete lists) but wrong for any pass that expresses deletes.

Failing test:
```kotlin
@Test
fun `transition count excludes deletes that were never applied`() {
    val repository = mockk<PropositionRepository>(relaxed = true)
    every { repository.query(any()) } returns emptyList()
    val report = DefaultDreamLoopOrchestrator.withRepository(repository)
        .withPass(FakePass("decay") { ConsolidationPassResult.Changed("decay",
            propositionsToDelete = listOf("doomed-1", "doomed-2")) })
        .consolidateNow(contextId)
    verify(exactly = 0) { repository.delete(any<String>()) }
    assertEquals(0, report.totalTransitioned)
}
```

dice/src/main/kotlin/com/embabel/dice/operations/consolidation/AbstractionPass.kt
```diff
@@ -0,0 +1,119 @@
...
+                    // SUPERSEDED with nothing to replace them — that would silently lose the facts.
+                    // Skip the group and leave the sources ACTIVE.
+                    skipped += props.size
+                    continue
+                }
+                toSave += abstractions
+                // withStatus preserves the contentRevised decay anchor (only metadataRevised moves).
+                // Pinned sources are eviction-immune: keep them ACTIVE alongside the abstraction
+                // rather than retiring them to SUPERSEDED.
+                toSave += props.filter { !it.pinned }.map { it.withStatus(PropositionStatus.SUPERSEDED) }
```
jasperblues, line 97: **A source mentioning two qualifying entities is superseded twice.** `distinct()` is applied only within a group, so a proposition in two qualifying groups is added to `propositionsToSave` as `SUPERSEDED` once per group (and fed to the abstractor in both). There's no `distinctBy { it.id }` on the aggregate — contrast `ContradictionResolutionPass`, which dedups. Also inflates `totalTransitioned`. (Same defect in `DefaultMemoryMaintenanceOrchestrator.abstract()`, whose comment asserts the opposite.)

Failing test (abbreviated — full version in the test file):
```kotlin
@Test
fun `a source mentioning two qualifying entities is superseded only once`() {
    // `shared` mentions entities "a" and "b"; groups a={a-1..a-4,shared}, b={b-1..b-4,shared}, threshold 5
    val result = AbstractionPass(abstractor, abstractionThreshold = 5).run(contextId, members)
    val changed = result as ConsolidationPassResult.Changed
    assertEquals(1, changed.propositionsToSave.count { it.id == "shared" })
}
```

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultCollectorRunner.kt
```diff
@@ -0,0 +1,248 @@
...
+                    skipped.addAll(propMarks)
+                    records += records(propMarks, runId, CollectorOutcome.SKIPPED, proposition.status, null)
+                }
+            }
+        }
+
+        // Compute the finish instant once and thread it into both the persisted run header and
+        // the returned result, so the audit object and the summary agree on the finish time.
+        val finishedAt = Instant.now()
+        persistRun(runId, startedAt, finishedAt, dryRun, records)
```
jasperblues, line 159: **A mid-run failure loses the audit trail for everything already mutated.** Mutations (`save`/`delete` + event emit) happen inside the loop, but `CollectorRecord`s are buffered and `persistRun` runs only after the loop, with no `try`/`catch`. If iteration N throws, iterations `1..N-1` are already committed but zero records are written — unrecoverable for `HARD_DELETED`. Suggest writing each record with (or before) its mutation, or wrapping so partial work is still recorded.

Failing test:
```kotlin
@Test
fun `a proposition transitioned before a mid-run failure still has an audit record`() {
    // two decayed candidates; first save() succeeds, second throws
    var saves = 0
    every { repository.save(any()) } answers {
        saves++; if (saves == 1) firstArg() else throw RuntimeException("storage failure")
    }
    runCatching { runner.run(contextId, dryRun = false) }
    assertEquals(1, listener.eventsOfType<PropositionStatusChanged>().size)   // first really transitioned
    assertTrue(recordStore.findByProposition(first.id).isNotEmpty())          // ...but no audit record
}
```

dice/src/main/kotlin/com/embabel/dice/proposition/revision/LlmPropositionReviser.kt
```diff
@@ -513,17 +513,28 @@ data class LlmPropositionReviser(
...
-                    .withStatus(PropositionStatus.CONTRADICTED)
-                    .copy(decay = acceleratedDecay, lastAccessed = Instant.now())
-                logger.debug(
-                    "Contradicted: {} (conf: {}, decay: {}) vs new: {}",
-                    original.text, reducedConfidence, acceleratedDecay, newProposition.text
-                )
+                // A pinned original is conflict-protected: a contradicting fact never auto-demotes
+                // it. Keep it untouched and store the new proposition alongside, leaving the
+                // conflict for explicit resolution.
+                val contradicted = if (original.pinned) {
```
jasperblues, line 519: **A contradiction against a pinned fact is silently swallowed — pinning has no "contested" signal.** Keeping the pinned original intact and storing the new fact alongside is the right call (conflict-protection, not suppression). But nothing durable records that the pin is now contradicted: this branch logs at `debug` and returns, and the consolidation twin (`ContradictionResolutionPass.kt:77`) emits *nothing at all* for a pinned weaker side. The KDoc says "left for explicit resolution," but there's no surfacing mechanism, so a wrong pin accumulates contradicting evidence invisibly and `unpin()` is never prompted. `ConflictClassificationGate.RouteToReview` already exists as vocabulary — emitting an observable outcome (event / `contested` marker) here would close the loop. Flagging as design rather than a test since it's a missing affordance, not a broken invariant.

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
```diff
@@ -127,12 +151,12 @@ class PropositionPipelineController(
         }
 
         // Process each chunk through the pipeline
-        val context = buildContext(contextId, emptyList(), schemaName)
+        val context = buildContext(contextId, parseKnownEntities(knownEntitiesJson), schemaName)
         val chunkResults = chunks.map { chunk ->
             val result = propositionPipeline.processChunk(chunk, context)
```
jasperblues: **The REST endpoints bypass `process()`, so the new resilience/concurrency layer is dead at this boundary.** Both handlers call `processChunk()` in a `chunks.map { }` loop instead of `process(chunks, context)`. Per the pipeline's own docs/tests, `processChunk` *propagates* failures while `process` isolates them — so `/extract/file` 500s the entire upload on a single bad chunk, the per-chunk `Failed`/`failedChunkIds` handling added here is unreachable, and the `Parallel`/`Batched` strategies (only consumed by `process()`) are never exercised via REST. Happy to add a MockMvc test (one failing chunk should yield partial results, not a 500) if you'd like.

jasperblues (commented):
Reviewed the maintenance + extraction subsystems closely (consolidation/dream-loop, mark-sweep reclamation, gates, the pinning lifecycle). Excellent structure overall — the pluggable-policy decomposition is clean and the test coverage is broad. A few findings cluster around one theme worth calling out: several "skip / protect / keep" branches drop information with no durable trace, which makes them hard to observe and, in two cases, means a feature silently doesn't run. Inline comments below; five come with a ready-to-run failing test (each encodes the intended contract, so it flips green once fixed). Happy to push these as a single `Pr48FindingsTest.kt` commit if useful — just say the word.

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
jimador: Good catch. Fixed so only a real cycle moves the baseline, growth accumulates across skips now instead of resetting every call.

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
jimador: Agreed. Added a reconcile step before the write, one copy per id with contradicted winning over superseded. Deterministic now regardless of pass order. Thanks.

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultDreamLoopOrchestrator.kt
jimador: Fixed. Only counts deletes when allowHardDelete is on, the same gate that applies them. Thanks.

dice/src/main/kotlin/com/embabel/dice/operations/consolidation/AbstractionPass.kt
jimador: Right on both. The pass dedupes its save list by id now and the maintenance twin tracks superseded ids across groups. Fixed the comment too. Thanks.

dice/src/main/kotlin/com/embabel/dice/projection/memory/DefaultCollectorRunner.kt
jimador: Good one. Wrapped the loop so partial work still gets recorded before rethrowing. Also found a sibling hole while testing, the record was buffered after the emit, so reordered to persist, record, emit. A throwing listener can't lose a record now. Thanks.

dice/src/main/kotlin/com/embabel/dice/proposition/revision/LlmPropositionReviser.kt
jimador: Agreed, both paths now emit `PropositionRoutedToReview` for a contested pin instead of going silent. Could make it a queryable marker in a followup if we want. Thanks.

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
jimador: Yep, switched /extract/file to process(). One bad chunk gives partial results now instead of 500ing the whole upload, and the parallel and batched strategies actually run over REST. Kept /extract on processChunk since it's a single chunk. Thanks.

jasperblues (approved):
Re-ran the seven findings against `1f1e8d2` — all addressed and verified green: the five correctness contracts (threshold accumulation, cross-pass reconciliation, transition counting, abstraction dedup, mid-run audit durability) now pass, the REST handlers route through `process()`, and pinned contradictions surface a `PropositionRoutedToReview` signal instead of being silently dropped. Clean, well-tested fixes. 🚀

PR #50, opened 2026-06-19
Metamodel + knowledge bundles

### LINKED ISSUE #45: Version the metamodel and catch drift before it corrupts the graph
Changes to the metamodel are silent. There's no way to see what changed between versions, and
mention-types that drift away from the model quietly land in the graph and corrupt it.

Version the metamodel so each version can be diffed against the last, and quarantine drifting types
for review instead of letting them silently through.

## Use cases

- See exactly what changed between two metamodel versions before adopting the new one.
- Catch a mention-type that has drifted from the model and quarantine it for review instead of
  letting it corrupt the graph.
- Roll the metamodel forward deliberately, with a record of each version to look back on.

### LINKED ISSUE #46: Snapshot and move what DICE knows
There's no way to move what DICE knows — the store is tied to the process and its backend, with no
portable, self-contained artifact to snapshot it or carry it elsewhere.

Add a local export/import format for knowledge bundles, so a graph — propositions, provenance, and
decay status — can be written out and reloaded faithfully in another instance.

## Use cases

- Snapshot a graph to a single portable file for backup.
- Carry knowledge from one DICE instance to another and reload it faithfully — propositions,
  provenance, and decay status intact.
- Share a curated knowledge set without handing someone access to the live store.

jimador:
## Metamodel + knowledge bundles

Last of the four-PR stack. **Stacked on #49** — review after #47–#49 merge; until then the diff
includes their commits.

Carves out two modules: `dice-metamodel`, `dice-bundle`.

### What's in it

- **Version the metamodel and catch drift before it corrupts the graph** (#45) — versions the
  metamodel so each version can be diffed against the last, and quarantines mention-types that drift
  from the model for review instead of letting them silently land in the graph.
- **Snapshot and move what DICE knows** (#46) — a local export/import format for knowledge bundles,
  so a graph (propositions, provenance, and decay status) can be written out and reloaded faithfully
  in another instance.

### Importer hardening

The Jackson bundle importer's parse path was tightened so untrusted input can't slip past the guards:

- A present-but-non-textual `formatVersion` is rejected through the version gate rather than coerced
  to the current version.
- Empty, whitespace, and `null` input returns a clean `ParseFailure` instead of throwing.
- The `maxBundleBytes` cap now applies to all three import paths (`importFromString`,
  `importFromStream`, `importFromReader`); the stream and reader paths are bounded mid-read, and the
  reader caps by UTF-8 byte length rather than character count.

No pre-existing design issues map to this PR.

Closes #45
Closes #46


jimador:
Closing — superseded by the incremental metamodel train. The metamodel content is being re-delivered in review-hardened slices on top of main; knowledge bundles will follow as their own series. Branch stays as source material.

PR #52, opened 2026-06-26
feat(mcp): expose DICE tools for MCP server integration

### LINKED ISSUE #5: Make it easy to expose an MCP server
Should probably just expose tools, perhaps with simplified parameters

LordKay-sudo:
## Summary

- Add `DiceMcpTools` with simplified `context_id`-scoped parameters for external MCP clients (Claude Desktop, Cursor, etc.).
- Add `dice-mcp-autoconfigure` to export those tools via embabel-agent `McpToolExport` when `embabel.dice.mcp.enabled=true`.
- Four tools only: `dice_recall`, `dice_list`, `dice_store`, `dice_get`. `get` refuses an id that belongs to another context.

Fixes #5

## Motivation

Issue #5 asks to expose DICE tools, perhaps with simplified parameters. External MCP clients are stateless, so every tool takes an explicit `context_id`. In-process `Memory` / `DiscoveryTools` still bake context in at construction; REST still takes it from the URL. MCP cannot do either, so isolation is a per-call check plus `PropositionQuery.forContextId`.

## Scope

This only exports `DiceMcpTools`. `DiscoveryTools` and `GraphQueryTools` stay on the in-process `asTools()` path. Extraction stays on `PropositionPipeline`.

## What changed

| Area | Change |
|------|--------|
| `com.embabel.dice.mcp` | `DiceMcpTools`, `DiceMcpSupport` |
| `dice-mcp-autoconfigure` | Opt-in Spring Boot wiring + named `diceMcpToolExport` bean |
| Docs | House-aligned README / architecture / AGENTS.md notes |

## Test plan

- [x] `DiceMcpToolsTest`: store/get, recall, list filters, context isolation, constructor guards
- [x] `DiceMcpAutoConfigurationTest`: opt-in, missing store, property bind, custom beans, ConditionalOnClass
- [x] full `mvn verify` (all modules, zero skipped tests)


LordKay-sudo:
Rebased onto latest main (through #60). Fixed the parent version to 0.1.1-SNAPSHOT. Build and DCO are green again. Ready when you are.

igordayen:
@LordKay-sudo - how are you? DICE is undergoing refactoring; could you please try to rebase? Thank you
looping @jimador

LordKay-sudo:
Hallo, I'm well thanks. Rebased onto current main and checked it against 0.2 / agent 1.5.

Same scope: context-scoped DiceMcpTools and opt-in autoconfig. Cut extract/assert and the core/extended profile so this stays four tools. Discovery/graph tools still out of this PR.

igordayen:
From CODEX:

 - P2 dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt:68: guarded() lets every IllegalArgumentException through. That includes exceptions thrown by repository.save/query/findById, not just caller validation. Example: if Neo4j/repository code throws IllegalArgumentException("MATCH ... bolt://neo4j-prod..."), that text is returned to the MCP client, despite the comment saying store details are sanitized. The tests only cover RuntimeException, so they miss this. Prefer a small internal validation exception, or validate inputs outside the guarded store block and sanitize all repository exceptions.

  - P2 dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt:187: tool names are fixed as dice_recall, dice_list, dice_store,
    dice_get, and dice-mcp-autoconfigure/.../DiceMcpAutoConfiguration.kt:82 exports them without a naming strategy. Embabel’s MCP
    export path deduplicates by final tool name, so if a host exposes another DICE tool set or another package also has dice_get, one tool can silently disappear, or the model cannot target the right one. Use a configurable/export-level prefix, e.g., embabel_dice_ or an app-specific prefix, via ToolObject naming strategy.

  - P3 dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpSupport.kt:111: "%.2f".format(...) uses the JVM default locale. Under a locale
    with comma decimals, the MCP output can become confidence=0,90 instead of confidence=0.90. Use Locale.ROOT so tool output stays stable across deployments.

igordayen (commented):
@LordKay-sudo - few comments from my side and codex. Please also be aware of PR 
https://github.com/embabel/embabel-agent/pull/1833 on tool name uniqueness. 
Please add comments to this PR, if you see better approaches, rather than Agent_Tool.
Thank you

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
```diff
@@ -0,0 +1,213 @@
...
+import com.embabel.dice.proposition.Proposition
+import com.embabel.dice.proposition.PropositionRepository
+import org.slf4j.LoggerFactory
+
+/**
+ * Simplified DICE tools for external MCP clients.
+ *
+ * In-process [com.embabel.dice.agent.Memory] and [com.embabel.dice.agent.DiscoveryTools] bake
+ * [ContextId] in at construction, so an agent cannot name another tenant. MCP clients are
+ * stateless and may serve many sessions, so every tool takes an explicit `context_id`. That is
```
jimador: I don't think stateless MCP requires the model to choose a context. With this shape, tenant selection becomes model-generated input and every host has to build authorization around it. Could we expose a host-supplied context policy, with a fixed-context implementation for the common case and an authorizing resolver for multi-tenant servers? Then the tool operates on a resolved `ContextId`, and caller-supplied context remains an explicit policy choice.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
```diff
@@ -0,0 +1,213 @@
...
+    @LlmTool(
+        name = RECALL,
+        description = "Search stored knowledge (propositions) in a DICE context. " +
+            "Pass a natural-language query to run hybrid semantic + keyword retrieval. " +
+            "Omit query to list memories ordered by confidence.",
+    )
+    fun recall(
+        @LlmTool.Param(description = "Context to search within (session, user, or tenant id).")
+        contextId: String,
+        @LlmTool.Param(description = "What to recall, in natural language. Omit to list all memories.")
```
jimador: The MCP schema marks `query`, `limit`, and `confidence` as required because `@LlmTool.Param.required` defaults to true and overrides Kotlin's optional-parameter metadata in `KotlinMethodTool`. That contradicts the descriptions and defaults. Please set `required = false` on all three and add an exported-callback test that inspects the input schema and invokes the tools with those fields omitted.

README.md
```diff
@@ -2412,6 +2412,51 @@ Everything is pushed into the database rather than scanned in memory:
 > `dice-storage/HANDOFF.md` for architecture and `dice-storage/INTEGRATE-INTO-ASSISTANT.md` for a
 > migration walkthrough.
 
+### MCP Server
+
+Expose DICE recall/list/store/get to an MCP client (Claude Desktop, Cursor, etc.) with
+`dice-mcp-autoconfigure` and embabel-agent's MCP server starter. Off until you set
+`embabel.dice.mcp.enabled=true`. Every tool takes a `context_id` — that is a scope, not a
```
jimador: The actual schema uses Kotlin parameter names verbatim: `contextId` and `propositionId`. There is no snake-case conversion in `KotlinMethodTool`, so a caller following this documentation sends arguments the tool does not bind. Please choose one public spelling, align the documentation with it, and test calls through the exported callback using the documented JSON.

dice-mcp-autoconfigure/src/main/kotlin/com/embabel/dice/mcp/autoconfigure/DiceMcpAutoConfiguration.kt
```diff
@@ -0,0 +1,84 @@
...
+        minConfidence = properties.minConfidence,
+        defaultLimit = properties.defaultLimit,
+    )
+
+    @Bean("diceMcpToolExport")
+    @ConditionalOnBean(DiceMcpTools::class)
+    @ConditionalOnMissingBean(name = ["diceMcpToolExport"])
+    fun diceMcpToolExport(tools: DiceMcpTools): McpToolExport {
+        logger.info("Exporting DICE MCP tools: {}", DiceMcpTools.TOOL_NAMES.sorted())
+        return McpToolExport.fromToolObject(ToolObject(objects = listOf(tools)))
```
jimador: This property exports read and write together. `dice_store` bypasses extraction, revision, and admission, then writes an ACTIVE proposition with caller-chosen confidence and no provenance. Retries can also create duplicate assertions. That is a materially different capability from recall. Please make writes independently opt-in, ideally through a host-replaceable assertion service that can apply provenance, admission, and idempotency policy.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpSupport.kt
```diff
@@ -0,0 +1,120 @@
...
+    /** Numbered rendering shared by `dice_recall` and `dice_list`. */
+    fun render(header: String, propositions: List<Proposition>): String =
+        buildString {
+            appendLine(header)
+            propositions.forEachIndexed { index, proposition ->
+                appendLine("${index + 1}. ${formatProposition(proposition)}")
+            }
+        }.trimEnd()
+
+    fun formatProposition(proposition: Proposition): String =
```
jimador, line 109: `dice_get` deliberately returns non-active propositions, but this shared summary omits status. A stale or contradicted fact therefore looks active to the model. Search and list can stay compact; get should use a separate detail contract containing status, effective confidence, and evidence references. I would make that an outward DTO now so we do not publish `Proposition` or lock MCP into this text format.

LordKay-sudo:
@igordayen `dice_recall`, `dice_list`, `dice_store`, and `dice_get` are a public contract. Qualifying them as `Agent_Tool` changes every MCP client for a collision that is two exports sharing a final name.

I would keep the `dice_*` names. If a host publishes more than one tool set, `ToolObject.withPrefix` is already on the 1.5 export path we compile against. That is the place for uniqueness. Collision should be reported, not solved by rewriting the name the client already uses.

`Agent_Tool` also gets hard to read once dots are sanitized. A host prefix stays something a person can type.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
LordKay-sudo: Agreed. A model-chosen tenant should not be the default. A host policy (fixed context for the common case, an authorizing resolver when the host is multi-tenant) is the right shape. The tools then run on a resolved `ContextId`.

Do you want that policy in this PR, or should I land the schema, sanitizer, and write-switch fixes first and do the policy next?

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
LordKay-sudo: `query`, `limit`, and `confidence` will be `required = false`. I will add an exported-callback test that checks the schema and invokes the tools with those fields omitted.

README.md
LordKay-sudo: Docs and tests will use `contextId` and `propositionId`, which is what `KotlinMethodTool` in `embabel-agent-api` 1.5.0-SNAPSHOT publishes. The callback test will send that JSON.

dice-mcp-autoconfigure/src/main/kotlin/com/embabel/dice/mcp/autoconfigure/DiceMcpAutoConfiguration.kt
LordKay-sudo: I will stop exporting `dice_store` unless `embabel.dice.mcp.writes-enabled=true`.

A replaceable assertion service (admission, provenance, idempotency) is a bigger SPI than this export. Say if you want that interface in this PR. I was going to treat it as a follow-up so a read-only host is the default without redesigning how facts are asserted.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpSupport.kt
LordKay-sudo: Search and list stay compact. `dice_get` will include `status`, effective confidence, and evidence refs so a stale or contradicted fact does not look active. That will go through a small detail type, not `Proposition`.

LordKay-sudo:
@igordayen The IAE hole is real. I will validate inputs outside `guarded` so a store `IllegalArgumentException` is sanitized. Locale will use `Locale.ROOT`.

`MethodTool` already logs the throwable. I will drop the second stack from `guarded`.

igordayen:
@LordKay-sudo - there are conflicts on the branch. Also, could you please mark it as "resolved" upon resolution? Thank you

LordKay-sudo:
@igordayen Rebased onto current main. Schema, writes, and get-detail threads are marked resolved. Left James's host context-policy thread open.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
igordayen: > Do you want that policy in this PR, or should I land the schema, sanitizer, and write-switch fixes first and do the policy next?

@LordKay-sudo - please assess the number of additional artifacts involved. This PR is already medium-size, thank you

LordKay-sudo: @igordayen Counted it. A host context policy is a new SPI, two implementations, autoconfig properties, schema changes on all four tools, a new test class plus rewrites of the isolation tests and the MCP docs. Callers currently have to pass `contextId`; a fixed-context host would take that off the model. That is a second public contract in this PR.

I would land this one as is and do the policy next.

igordayen (approved):
@LordKay-sudo - looks good; thank you for contributing!

LordKay-sudo:
@igordayen Thanks.

dice/src/main/kotlin/com/embabel/dice/mcp/DiceMcpTools.kt
jimador: Thanks @LordKay-sudo, that’s fair. Let’s keep this scoped and follow up on context policy once we have a concrete client integration to drive it. My concern about model-selected context still stands, but I don’t think we need to design the full abstraction here.

PR #53, opened 2026-06-29
Converge the vector index name on one canonical source

jimador:
## One vector index name, not three

Three paths resolved the proposition vector index name independently and could disagree, so vector
search silently returned nothing. The annotation-bound search path always used
`Proposition_embedding_vector`, but a blank `vector-index.name` derived that name on the `findClusters`
side while flowing through as `""` to the schema (DDL) side — so the DDL never created the index the
query looked for. A custom `name`, or a changed `label`/`property`, split them the same way.

The fix makes the index identity non-configurable instead of configurable-but-validated. The
`@VectorIndex` annotation on `PropositionNode.embedding` is the single source of truth:

- `DrivinePropositionRepository` holds the canonical constants — `VECTOR_INDEX_LABEL`,
  `VECTOR_INDEX_PROPERTY`, and `VECTOR_INDEX` (derived from them).
- The schema (DDL) spec and the `findClusters` query both wire to those constants, so all three paths
  hit the same index by construction — there's nothing to keep in sync and nothing to validate.
- `vector-index.name`, `label`, `property`, and `similarity-function` are dropped from config — they
  were knobs that could only ever hold their annotation-bound defaults. `vector-index.enabled` stays;
  it's the one that genuinely varies.
- Regression test (pure JVM, no Neo4j) asserts the constants match the annotation and that the schema,
  `findClusters`, and `loadNearest` names all agree.

### Why the index name isn't configurable

dice resolves the vector index name from the `@VectorIndex` annotation, not config. Every
`loadNearest` / `findSimilar*` path routes through Drivine's `VectorIndexResolver`, for any caller.
The only override is the annotation's own `name` attribute, a source change — so a Spring property
here would silently fail to redirect search the way you'd expect.

Nothing consumes this yet (pre-release), so no migration needed. Note for later: drivine matches an
existing index by `(label, property)` and ignores its name, so a DB carrying a legacy differently-named
index needs it dropped (data-safe — embeddings are node properties, the index rebuilds) before the
canonical one appears.


jasperblues:
Looks good 👍

jasperblues (approved):

PR #57, opened 2026-07-01
feat(dice): dedup grounding merge, multi-signal collector, native graph queries (#56)

### LINKED ISSUE #56: Dedup sweep drops grounding/provenance when collapsing duplicates (pure STALE flip, no merge onto survivor)
## Summary

When the **dream-loop dedup sweep** collapses a cluster of near-duplicate propositions, the losing
duplicates are transitioned to `STALE` with a **pure status flip** — their `grounding`,
`provenanceEntries`, `sourceIds`, and `reinforceCount` are **not** merged onto the surviving
proposition. Because `STALE` is excluded from standard retrieval, all of that evidence becomes
invisible. Collapsing a duplicate therefore *loses* corroboration instead of consolidating it: if a
lower-confidence loser was grounded in three sources and the survivor in one, the graph ends up
citing one source where it should cite four.

This is scoped specifically to the dedup sweep path:
`DuplicateCollectorStrategy` → `StatusTransitionSweepPolicy` → `DefaultCollectorRunner`.

## Evidence (verified against `main`)

1. **Pure status flip on collapse** — `DefaultCollectorRunner.kt:121`:
   ```kotlin
   val saved = repository.save(proposition.withStatus(action.newStatus))
   ```
   This is the only mutation on the `TransitionStatus` branch; nothing else is copied.

2. **`withStatus` copies only status + timestamp** — `Proposition.kt:255-256`:
   ```kotlin
   fun withStatus(newStatus: PropositionStatus): Proposition =
       copy(status = newStatus, metadataRevised = Instant.now())
   ```

3. **Survivor is chosen, loser is marked with the survivor id** — `DuplicateCollectorStrategy.kt:89`
   picks `maxWith(effectiveConfidence, reinforceCount, id)` and marks each loser with
   `MarkReason.Duplicate(survivorId = survivor.id)` (`:95-98`).

4. **`survivorId` is audit-only** — every read of `survivorId` flows into the collector audit record
   (`MarkReason.kt:44`, `LineageRowMappers.kt:73/86`, `DrivineCollectorRecordStore.kt:50`). No code
   path reads it to re-point or merge data onto the survivor.

5. **The policy structurally cannot merge** — `SweepPolicy.decide(proposition, marks)`
   (`SweepPolicy.kt:31-41`) receives no repository handle and never inspects
   `MarkReason.Duplicate.survivorId`; `StatusTransitionSweepPolicy` just returns
   `TransitionStatus(STALE)` (`SweepPolicy.kt:62-67`).

6. **The loser's evidence goes invisible** — `STALE` is excluded from retrieval
   (`Memory.kt:186-192`, `GraphQuery.kt:152`; enum doc `Proposition.kt:50-52`).

## Precedent — dice already knows how to do this

The **ingestion-time** merge already folds grounding onto the survivor. `LlmPropositionReviser`:
```kotlin
val combinedGrounding   = (existing.grounding + new.grounding).distinct()
val combinedProvenance  = (existing.provenanceEntries + new.provenanceEntries).distinct()
existing.copy(grounding = combinedGrounding, provenanceEntries = combinedProvenance,
              reinforceCount = existing.reinforceCount + 1, ...)
```
(`LlmPropositionReviser.kt:666-708`, in `mergePropositions` / `reinforceProposition`.) And the copy
helpers already exist on `Proposition` — `withGrounding` (`:276`), `withProvenanceEntries` (`:288`),
`withProvenance` (`:294`) — they are simply never invoked on the sweep path. So this is a wiring gap,
not a missing capability.

## Why it matters

- **Evidence loss.** The "why do I believe this" chain is silently truncated on every collapse.
- **Consumers depend on grounding.** Anything that grounds output in cited proposition chains (e.g. a
  consequence/coverage layer) is starved by collapse.
- **The claim "collapse strengthens grounding" is currently false** — collapse *weakens* it.

## Proposed fix (direction, not mandate)

The policy SPI has no repo handle, so the merge belongs in the runner. Suggested shape, consistent
with the existing "policy decides intent, runner does IO" split:

1. Add a `SweepAction.MergeInto(survivorId: String, thenStatus: PropositionStatus = STALE)` variant.
2. Add a `MergingSweepPolicy` that reads `MarkReason.Duplicate.survivorId` from the marks and emits
   `MergeInto(...)` (falls back to `TransitionStatus(STALE)` when there is no survivor id).
3. In `DefaultCollectorRunner`, handle `MergeInto` by loading the survivor, folding the loser's
   `grounding` / `provenanceEntries` / `sourceIds` onto it (reuse `withGrounding` /
   `withProvenanceEntries`; union + distinct) and bumping `reinforceCount`, saving the survivor, then
   transitioning the loser to `thenStatus`. Emit the existing status-changed event for the loser.

Keep `StatusTransitionSweepPolicy` as-is; select the merging policy where the dedup sweep is
assembled. Behaviour must remain non-destructive and dry-run-safe (dry run reports, mutates nothing).

## Scope / non-goals

- **In scope:** the dream-loop dedup sweep path only.
- **Not this issue:** the ingestion-time reviser merge (already correct); `AbstractionPass` /
  `ContradictionResolutionPass` (separate SUPERSEDED/CONTRADICTED flips — flag if they warrant the
  same treatment, but out of scope here); entity resolution.

## Acceptance criteria

- A test proves the **current** defect: collapse a cluster where a loser carries grounding/provenance
  the survivor lacks → survivor is unchanged (documents the bug), loser is STALE.
- After the fix: the survivor gains the union of the losers' `grounding` + `provenanceEntries`
  (deduped) and a bumped `reinforceCount`; the loser is STALE; retrieval reaches every original
  source via the survivor.
- Dry-run mutates nothing.
- Follows existing collector test patterns (`DefaultCollectorRunnerTest`, the in-memory repo +
  `FixedVectorEmbeddingService` integration harness).

jimador:
Closes #56.

This PR delivers three features: duplicate collapse that preserves evidence, a multi-signal duplicate detector with a full audit trail, and native graph queries with server-side filtering for the Neo4j backend.

## Evidence-preserving duplicate collapse

Previously the sweep collapsed near-duplicate propositions by flipping the losers to STALE. STALE propositions are excluded from retrieval, so every loser's grounding and provenance disappeared with it. Now the loser's evidence folds onto the survivor before the loser is retired.

- `SweepAction.MergeInto(survivorId, thenStatus)` is the new collapse outcome. `MergingSweepPolicy` produces it from a `MarkReason.Duplicate` mark and falls back to plain retirement when a mark is pinned, unmarked, or names no usable survivor.
- `DefaultCollectorRunner` folds grounding, provenance, and source ids onto the survivor (append + distinct), increments `reinforceCount`, saves, then retires the loser with the same status-changed event as before. Dry runs mutate nothing.
- When merges chain within one run (X merges into A while A merges into B), targets resolve to the terminal survivor, with deterministic cycle breaking, so evidence never lands on a proposition retired in the same run.
- `MergingSweepPolicy` is now the builder default. Any `CollectorRunner` built without an explicit `withPolicy()` gets merge-on-duplicate semantics; `withPolicy(StatusTransitionSweepPolicy())` restores pure status flips. This is a deliberate behavior change for existing callers and the main thing to weigh in review.

Adopter note: `save()`'s duplicate check assumes a `(contextId, text)` uniqueness constraint as a cross-instance backstop. Apply it in production schema.

## Multi-signal duplicate detection

Cosine similarity alone misses duplicates that reword a fact and, worse, can merge contradictions ("Alice works at Acme" / "Alice no longer works at Acme" are near-identical by embedding distance). `MultiSignalCollectorStrategy` replaces the single cosine cut with a pipeline: propose candidate pairs, score each pair on independent signals, blend the scores into one edge, group edges into connected components, pick one survivor per component, and mark the rest as duplicates.

- Extension interfaces in `spi`: `CandidatePairSource`, `CollectorSignalScorer` (null return abstains), `CollectorTraceStore`, `CollectorTraceQuery`, `ConnectedComponentsFinder`. Every collaborator is replaceable.
- Six built-in scorers: vector similarity, lexical similarity (commons-text JaroWinkler), entity overlap, grounding overlap, provenance overlap, and a polarity veto that blocks contradictory pairs from merging. The overlap scorers use a set-based Jaccard over domain ids; a code comment documents why commons-text's character-level `JaccardSimilarity` is not a substitute.
- Contradictions cannot merge transitively: after components form, every intra-component pair is checked, and pairs vetoed at the edge stage count as standing contradictions, so A and C never merge through a shared neighbor B.
- Component formation is an in-JVM union-find. No GDS (GPLv3), no APOC; runs on any Neo4j.

## Collector audit trail

Every run records what it saw and what it decided, keyed by run id: candidate edges with their per-signal scores (including vetoed edges, so the trace answers why two propositions did *not* merge), components, and decisions carrying the retired members and their folded evidence references — enough to explain or reverse a collapse after the fact.

- `DrivineCollectorTraceStore` (Neo4j, plain UNWIND Cypher) and `InMemoryCollectorTraceStore`. Reads are corrupt-row tolerant: a row that fails to map is skipped and logged, never fails the whole read. `findDecisionsByRun` fetches a run's decisions and their retired members in a single query.
- Trace writes are best-effort. A trace-store failure logs a warning; the run's marks and merges proceed.
- Spring auto-configuration wires the store and the collector; every signal weight and threshold is tunable under `embabel.dice.collector.*`. The store bean serves both `CollectorTraceStore` and `CollectorTraceQuery`.

## Native graph queries and server-side filtering

`DrivinePropositionRepository` now implements `GraphQueryCapable`: neighborhood, path, and lineage queries run as native Cypher (bridging hops through `resolvedId`, since propositions have no entity node), including context-scoped queries via the `honorsContextFilter` opt-in that confines every hop to the query's context. `GraphQuery` routes to the native path when the store declares the capability; a parity integration test pins native results to the portable implementation and asserts no cross-context leakage. The caller's `maxDepth` is honored end to end and clamps at a documented ceiling with a warning rather than truncating silently.

Filtering and counting that previously ran on the JVM now run in the store: `findAbstractionsOf`, decay-status filtering, `count(query)`, `keywordOverlap`, `belowEffectiveConfidence`, and the decay sweep's `effectiveConfidence` recompute (a Cypher CASE expression batched with `CALL {} IN TRANSACTIONS`). dice is a general library; it cannot assume a consumer's data fits in memory.

New public API, all additive with default implementations: `PropositionStore.count(query)` and `keywordOverlap`, `PropositionQuery.belowEffectiveConfidence`, the `GraphQueryCapable` context and depth overloads, and `CandidateSearcher.resolutionLevel`.

## Docs

`docs/design/` audited against source: signature and enum drift corrected, duplicated prose removed, code examples added, and `multi-signal-collector.md` / `collector-trace-store.md` describe the shipped behavior, including the sweep-policy default change.


dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
```diff
@@ -0,0 +1,289 @@
...
+
+    @Transactional
+    override fun recordCandidateEdges(runId: String, edges: List<CollectorCandidateEdge>) {
+        if (edges.isEmpty()) return
+        logger.debug("Recording {} collector candidate edge(s) for run {}", edges.size, runId.take(8))
+
+        val edgeRows = edges.map { CollectorCandidateEdgeRowMapper.edgeBindMap(runId, it) }
+        persistenceManager.execute(
+            QuerySpecification.withStatement(
+                """
```
jasperblues, line 83: If desired you can also inject queries that are kept in separate files. (These are small so maybe they are better inline).

jasperblues (approved):

Issue #64, opened 2026-07-24
Carry opaque source revisions in proposition provenance


jimador:
## Goal

Carry an opaque source revision through DICE extraction and existing source provenance so
successive versions of one logical source are queryable without changing locator identity.

## Context

DICE already records typed source locators and `DerivedFrom` grounding. Source revision is
an additional scalar property of that source relationship, separate from extraction-run
lineage, and must not add a new `SourceLocator` subtype or duplicate the locator key.

The design follows stable-key + opaque-version APIs such as S3, W3C PROV's stable
entity/specialization split, and OpenTelemetry's distinction between one parent and many
causal links. `SourceRevisionRef(sourceKey, sourceRevision)` is the exact-query/run-lineage
value; stored provenance keeps only the scalar revision beside its existing locator.

## Acceptance criteria

- [ ] Extraction entry points accept an optional opaque source revision.
- [ ] Existing source-grounding equality and locator keys remain stable.
- [ ] `SourceRevisionRef.sourceKey` equals canonical `SourceLocator.key()` at the DICE boundary.
- [ ] Store APIs provide tenant-scoped all-version (including revisionless), exact-version,
      and revisionless-only queries with Drivine push-down.
- [ ] Legacy revisionless callers remain compatible.
- [ ] REST and canonical proposition JSON round-trip source revision losslessly.
- [ ] Two revisions of one source remain related and independently queryable.
- [ ] Tests prove source grounding does not multiply when two runs analyze one revision.
- [ ] After #46 lands, real unit and Neo4j knowledge-bundle round-trips preserve revision;
      this issue remains open until that external gate passes or acceptance is revised.

## Constraints

- Keep source revision independent from extraction-run lineage.
- Do not infer provider revision semantics inside DICE.
- One aggregate input may carry one truthful aggregate revision; #67 owns the ordered set
  of causal member source revisions.
- Coordinate bundle changes with DICE issue #46 and metamodel versioning with #45.
- The operator accepted source/JSON/Java compatibility and existing Java constructor
  descriptors as the boundary; full Kotlin synthetic constructor/copy ABI is not claimed.
- PR #71 completed the Spring/Embabel upgrade; the operator confirmed `main` is the sole
  delivery target and no backport is required.
- Before the PR, disposable copies of the checked-out Embabel `assistant` and `me` apps must
  compile and pass representative DICE-using tests against the integrated `0.2.0-SNAPSHOT`;
[...]

PR #73, opened 2026-07-28
Carry source revisions in proposition provenance

jimador:
## Summary

Carry an optional opaque source revision through extraction, stored provenance, tenant-scoped queries, REST/discovery surfaces, collector fold/undo, and JSON without changing source locator identity.

Source revision is intentionally distinct from extraction-run lineage: `SourceRevisionRef(sourceKey, sourceRevision)` identifies one logical source version, while run metadata continues to identify the analysis execution.

## Compatibility boundary

- Preserves existing Kotlin and Java source compatibility.
- Preserves the accepted old Java constructor arities and existing public method descriptors with executable interop checks.
- Keeps legacy JSON readable and makes source revision additive; existing public trace locator values remain readable and retain their legacy meaning.
- Preserves stored revisionless provenance and lazily adopts exact legacy graph edges when revisionless data is rewritten.
- Does **not** claim full Kotlin synthetic constructor/copy binary ABI compatibility; that boundary remains explicitly out of scope.

## Behavior

- Adds tenant-scoped queries for all source versions, an exact revision, and revisionless-only provenance.
- Keeps `SourceLocator.key()` stable and stores revision as a scalar on existing source evidence.
- Uses revision-distinct Drivine relationship identity, with exact legacy-edge adoption and retry-safe race recovery.
- Preserves revision-distinct evidence through collector fold/undo while retaining legacy trace values for consumers.
- Round-trips source revision through canonical proposition JSON, REST, and discovery DTOs.

## Review fixes

The issue was that a few edge paths could reuse the wrong source, partially mutate an undo, or leave compatibility and query-plan behavior asserted indirectly. The fix validates source identity before dedup reuse, resolves both undo participants before writing, proves persisted replay cardinality, pins an executable old-client binary fixture, and couples plan checks to the production query invocation.

## Validation

- `./mvnw verify` — all seven Dice modules passed (1,246 Dice tests and 110 storage tests).
- Me downstream extraction, ingestion, storage-guard, and unmerge suite — 40 passed.
- Live Me startup from `.env` succeeded; health is UP and James's read-only Discord handler returned HTTP 200 with an empty result using a future cursor.
- No database migration is required.
- No experimental API markers are added.

Refs #64. The remaining bundle integration gate waits on #46 and is not required for this PR's core source-revision behavior.


igordayen:
@jimador - good development.  would be helpful if you could add a diagram (or informal text) of artifacts using SourceRevisionRef and the relationships between those artifacts. I see approx. 5-6 of those artifacts. A bit challenging to understand the full picture, as PR is sizable. 
thank you

jimador:
Closing — this will be re-sliced onto main as incremental PRs as part of the 1.0-M1 train. Branch stays as source material.

PR #76, opened 2026-08-02
Upgrade to jackson3

guanxuc:


igordayen (approved):
looks good, @guanxuc  thank you

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
```diff
@@ -33,15 +33,15 @@ import com.embabel.dice.pipeline.ChunkPropositionResult
...
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
-import com.fasterxml.jackson.module.kotlin.readValue
 import org.slf4j.LoggerFactory
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.*
 import org.springframework.web.multipart.MultipartFile
+import tools.jackson.databind.ObjectMapper
```
jimador, line 42: Can we verify that existing consumers can still start and use Dice JSON with Jackson 2 on the classpath? If not, this needs a compatibility period before changing the public ObjectMapper type.

dice/src/test/kotlin/com/embabel/dice/proposition/PropositionJacksonRoundTripTest.kt
```diff
@@ -32,7 +29,7 @@ import org.junit.jupiter.api.Test
  */
 class PropositionJacksonRoundTripTest {
 
-    private val mapper = jacksonObjectMapper().findAndRegisterModules()
+    private val mapper = jacksonObjectMapper()
```
jimador, line 32: Can we add a JSON fixture written by the Jackson 2 repository and load it here? This only tests a Jackson 3 round trip, so it does not cover existing persisted data.

jimador (changes requested):
Looks great! Thank you. Just a couple of house keeping comments for existing clients.

jimador:
Carrying a note over from the review of #84: `dice-storage/src/main/kotlin/com/embabel/dice/storage/MetamodelRowMappers.kt` builds its own `com.fasterxml.jackson.databind.ObjectMapper` as a private top-level value.

`EmbabelObjectMapperHolder` is the framework's holder and is already on `tools.jackson`, so this mapper should move to it as part of this upgrade. Wiring it before the upgrade would pull Jackson 3 into `dice-storage` on its own.

PR #82, opened 2026-08-30
docs(specs): competitive positioning + developer docs architecture

jimador:
**Breaking changes: none.** Specs only; no code, no API.

Two specs, both rewritten in this round.

- `specs/competitive-positioning.md`: what DICE is against Zep/Graphiti, Mem0, Letta, Cognee, Hindsight, LangMem, Neo4j agent memory, and the three managed memory services, on the unit of knowledge, projections, knowledge hygiene, schema governance, shared mechanisms, remaining gaps, candidate modules, and orthogonal research. Every claim derives from what `main` states (README, `docs/design`) with anchored citations; material that only exists on unmerged branches was removed, and four claims that were wrong against the code were corrected (valid-time fields exist, so the gap is transaction time; bi-temporal is held by Zep; the pipeline is not serial and pinning exists; the +0.15/×0.7/×0.85 figures are decay adjustments). Open questions each carry a recommendation.
- `specs/readme-to-docs-tree.md` (replaces `specs/developer-docs-architecture.md`): the `docs/` layout, which README section (by line range) lands on which page, the concept-page order, the feature-page template with each opt-in surface's activation condition, the quickstart steps and constraints, the compatibility matrix axes, and the rule that every feature PR ships its docs page.


jimador:
This is not ready to be the positioning and docs authority yet. Several central claims disagree with the current code or current product documentation:

- `competitive-positioning.md:183-338` mixes product position with design status, current gaps, brief numbers, and candidate implementation work. Move delivery and planning into roadmap/research documents; keep this document derived from the proposition-first product intent.
- `competitive-positioning.md:54-55,291-293` says DICE cannot represent world-valid time. `TemporalMetadata` already carries observed time, `validFrom`, `validTo`, invalidation, and ingestion time. The missing transaction-history/query behavior is narrower.
- `competitive-positioning.md:318-319` says extraction is serial and consumers batch manually. `ExtractionExecutionStrategy` already has parallel and batched implementations; entity resolution is the serial stage.
- `competitive-positioning.md:114-115,262-265` says surveyed systems expose no hard TTL deletion or TTL API. Current [Google](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/agent-engine/memory-bank/set-up), [AWS](https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_Memory.html), and [Microsoft](https://learn.microsoft.com/en-us/azure/foundry/agents/how-to/memory-usage) documentation contradicts that comparison.
- `developer-docs-architecture.md:130-135` proposes dropping vector retrieval to avoid configuring embeddings, but the in-memory autoconfiguration eagerly obtains the default embedding service. Boot the exact quickstart without an embedding model before documenting this path.
- `developer-docs-architecture.md:52` calls the REST API API-key-gated. `DiceRestConfiguration` imports the controllers only; API-key security is separate and disabled by default. The feature page needs the actual activation contract, or REST activation needs to fail closed.
- `competitive-positioning.md:99-101` presents one pass threshold and revival path as a system invariant. The shipped transition policies use different thresholds and behavior. Keep the exact rule in one lifecycle authority and describe this as policy-configured here.

There are smaller authority problems too: graph/Prolog ownership at lines 67-69, existing authority/trust/pinning/execution capabilities listed as missing at lines 313-319, Oracle classified as a materialized projection, and a review-blocking docs policy defined only in an internal strategy file. These should be corrected in the same pass so the new documents do not become another conflicting copy of the architecture.

jimador:
Three more things on top of the thread above, from reading this against the two open stacks.

Delivery status will be wrong on the day the stacks land. `competitive-positioning.md:183-185` says the stamp, drift detection and quarantine are design-stage, and `:291` lists the extraction-run identifier as missing. #83 to #88 and #95 to #101 change both statements. Keep a shipped / in delivery / planned table keyed on PR numbers and require each feature PR to move its own row. Prose status goes stale silently.

The vocabulary covers one stack. Drift, contract and quarantine map onto `DriftReport`, `DeclaredSchema` and `QuarantineDecision`. Source revision, extraction profile, extraction run and run lineage appear nowhere except the missing bullet at `:291`. Those are the types the other nine PRs add, and the doc needs a section that positions them: stable key plus opaque revision, the run as the unit of attribution, the profile as host-owned policy identity.

Quarantine is described as the disposition. The shipped default is `observe`, a scheduled check never quarantines, and there is no read or release surface (comment on #88). Describe it as opt-in behind an explicit live call, and say that declared renames never rewrite stored data.

The per-PR docs contract at `developer-docs-architecture.md:264-282` fails all fifteen open PRs, since none ships a developer page and the `docs/` tree does not exist yet. Decide before merge order is set. I would grandfather the in-flight stacks and land one docs PR per stack written from the design notes.

specs/competitive-positioning.md
```diff
@@ -1,198 +1,271 @@
...
+| **LangMem** | Semantic, episodic and procedural memory items | Optional typed Pydantic profiles; consolidation state |
+| **Hindsight** | Structured facts in a knowledge graph | Entity resolution that links "Alice" to "my coworker Alice" |
+| **Google / AWS / Microsoft** | Extracted memories as flat fact strings or items | A strategy or topic label and an IAM scope |
+| **DICE** | Proposition | Confidence, importance, decay, status, grounding, provenance entries, reinforce count |
+
+Three of those units carry a field DICE's proposition lacks.
+
+- **Bi-temporal edges (Zep / Graphiti).** World time is stored separately from ingestion time.
+  DICE's `TemporalMetadata` carries `observedAt`, `validFrom`, `validTo` and `invalidatedAt`, with
+  no transaction-time axis, so a query for what the store believed at a past instant has no basis.
```
igordayen, line 53: no historical dimension, correct understanding?

specs/competitive-positioning.md
```diff
@@ -1,198 +1,271 @@
...
+- **Ontology grounding (Cognee).** Extracted entity and type names resolve against a declared
+  OWL/RDF ontology, with fuzzy matching, before any graph node is built
+  (https://docs.cognee.ai/core-concepts/ontologies). DICE declares types in a `DataDictionary` and
+  applies them at extraction, with no resolution against a stored ontology.
+- **Change history (Mem0).** Old value, new value, event and actor per change. DICE records why a
```
igordayen, line 58: old / new value is an audit aspect, correct understanding?

specs/competitive-positioning.md
```diff
@@ -1,198 +1,271 @@
...
-
-**Their moat**: Excellent extraction prompts, prompt optimization via gradient analogy, debounced background reflection, dilated windows retrieval.
+| Operational tooling and memory inspection | 4 | Bedrock emits CloudWatch metrics for latency, invocations, errors and memory creation count, with spans over CreateEvent, GetEvent, ListEvents, DeleteEvent and RetrieveMemoryRecords, plus extraction and consolidation logs ([metrics](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-memory-metrics.html)). Vertex lists stored memories in the Cloud Console (Memory Bank UI announcement, Google Developer Forums, no stable URL recorded). The NAMS dashboard shows health, entity count and queue lag with graph visualisation and Cypher queryability ([tour](https://medium.com/neo4j/a-tour-of-the-neo4j-agent-memory-service-nams-0f2d535a4fdb), 2026-06, GA unconfirmed). LangSmith traces execution paths and state transitions with cost, latency and error dashboards (langchain.com/resources/llm-observability-tools) | "Why was this proposition formed" needs custom logging over `CollectorTraceStore` |
```
igordayen, line 171: addition to observability?

specs/competitive-positioning.md
```diff
@@ -1,198 +1,271 @@
...
+AGM's postulates cover logical theories and DICE's beliefs are graded, so AGM serves as a checklist.
+LLM disambiguation already covers what Fellegi-Sunter scoring would add. The first two opportunities
+touch `Proposition.kt` and `ContradictionResolutionPass.kt`, which consolidation, projection and
+retrieval ranking all depend on.
+
+## Open questions
+
+- **Which working-memory capabilities belong in DICE and which in the consuming application.**
+  Recommendation: activation ranking and per-turn lifecycle events in DICE, prompt-region and budget
+  policy in the application, since only the application knows its token budget.
```
igordayen: There is budget abstraction in ```embabel-agent```.  
Token consumption and budget are available in tool loop interceptors and observability APIs.

specs/competitive-positioning.md
```diff
@@ -1,198 +1,271 @@
...
+  (`docs/design/architecture.md`).
+
+## Out of scope
+
+- **Zep's five-reranker retrieval.** Deep feature tied to Neo4j traversal. The bi-temporal model is
+  the better investment from that system.
+- **LangMem's prompt optimisation.** Orthogonal to memory quality.
+- **Mem0's separate graph pipeline.** DICE has entity mentions plus a Neo4j projection over the same
+  propositions.
+- **Google's multimodal extraction.** The proposition model is format-agnostic. Add on demand.
```
igordayen: There is already an issue on multimodal support for LLM output. Input is supported already

specs/readme-to-docs-tree.md
```diff
@@ -0,0 +1,140 @@
+# Split the README into a docs tree
+
+Decides the `docs/` layout, which README section lands on which page, the shape of the quickstart
+and feature pages, and the rule that keeps pages current. Writing the pages is out of scope.
+
+## Target tree
+
+| Directory | What goes there |
+|---|---|
+| `docs/README.md` | What DICE is, when to use it, links into the tree. |
```
igordayen: very nice!

specs/readme-to-docs-tree.md
```diff
@@ -0,0 +1,140 @@
...
+
+Every feature PR ships its developer-doc page and its design-doc delta. Consumer-visible behaviour
+updates the page under `docs/`, new rationale updates the `docs/design/` note, an opt-in feature
+gets a feature page carrying its activation condition, a new property updates
+`docs/reference/configuration-properties.md`, and a version-support change updates the compatibility
+matrix. Internal refactors with no consumer-visible change, test-only changes and build changes are
+exempt. Reviewers ask one question: with only this PR's docs, could a consumer use the feature? A no
+is a blocking finding. A promise of later docs does not clear it.
+
+## Open questions
```
igordayen, line 133: Is AsciiDoc considered for documentation as well?

igordayen (approved):
@jimador - very nice writeup. few non-blocking inquiries. thank you

specs/competitive-positioning.md
jimador: Sort of. The valid-time axis is there: `validFrom`, `validTo` and `invalidatedAt` on `TemporalMetadata`, plus `supersedes` and `contradicts`, so "when was this true in the world" is answerable.

What is missing is transaction time, the instant DICE recorded a belief or changed one. `Proposition` carries `created`, `contentRevised` and `metadataRevised`, but those describe the current row only. No prior version is retained, so the belief set as of a past instant cannot be reconstructed and a superseded value's prior state is not queryable. Zep and Graphiti keep both axes on the edge, which is what the bullet points at.

Worth capturing, so I will open an issue for it.

specs/competitive-positioning.md
jimador: Yes, it is an audit concern. Most of it is in flight: #95 gives a run a pseudonymous `ExtractionActorRef`, #101 links each proposition to the run that produced it, and #92 and #103 record a fold with the retired proposition's prior status so it can be undone. What is left is an old and new value per field for confidence and status changes, which nothing open covers.

specs/competitive-positioning.md
jimador: Yes, and it splits in two. The LLM half already rides on the framework: `AgenticEventListener` with `LlmInvocationEvent` gives token usage, cost and model per extraction call, and the observation contexts cover spans. The half DICE owns is memory-specific: why a proposition was formed, what a sweep merged or retired, how much decayed. `CollectorTraceStore` holds that data and #103 records the decisions and exposes an unmerge over them, but there are no metrics and no inspection surface. #8 lifecycle events would carry the counters.

specs/competitive-positioning.md
jimador: Correct, and the reason given in the doc is wrong. `Budget` in `com.embabel.agent.core` carries cost, action and token caps on `ProcessOptions` and turns them into an `EarlyTerminationPolicy`, with `TokenBudgetGuardRail` on input and consumption visible through `ToolLoopInspector` and the invocation events. Budget policy stays outside DICE because the framework owns it, not because the application has to build it. Reworded the recommendation.

specs/competitive-positioning.md
jimador: Right. Input landed upstream in embabel/embabel-agent#43 and `MultimodalContent` carries images and documents. Output is still open as embabel/embabel-agent#42. The line stays out of scope for DICE, but the reason is that the input path already exists and the proposition model is format-agnostic, so wiring it is small when a case shows up. Updated.

specs/readme-to-docs-tree.md
jimador: Thanks. That tree is what the docs rule at the bottom enforces, so a feature PR knows which page it owns before it starts.

specs/readme-to-docs-tree.md
jimador: Not yet, and it should be. DICE has no `.adoc` today and the `docs/` tree is 20 Markdown files, while `embabel-agent` publishes its reference from `embabel-agent-docs/src/main/asciidoc` through the asciidoctor Maven plugin to HTML and PDF. My take: Markdown while the tree is read on GitHub, AsciiDoc when DICE docs are meant to publish into the same site as the framework reference, since that is where the toolchain and the includes already live. Naming AsciiDoc in the open question so the fork is explicit.

jimador: Settled the other way, and better: DICE docs follow the framework and use AsciiDoc, built the way `embabel-agent-docs` is. The page paths through the spec are now `.adoc`, GitHub renders them in the repo before any site exists, and the existing Markdown under `docs/design/` converts as those notes get touched. Whether DICE renders its own site or publishes into the framework's stays open.

specs/competitive-positioning.md
jimador: Correcting myself: "no metrics and no inspection surface" is wrong. `ObservableMentionFilter` already records Micrometer counters, `DiceEventListener` carries the lifecycle taxonomy including `PropositionStatusChanged(previousStatus, newStatus, reason)`, `ExtractionBatchCompleted` and `ProjectionBatchCompleted`, and `DiscoveryController` serves `/why/{propositionId}` and `/projection-health`. The hooks are there and the host owns the `MeterRegistry`, which is the right split for an embeddable library.

What is missing is the documentation, and this PR already places it: `docs/production/` owns the observability page in the docs tree. So no tracking issue. The one real gap from this review is durable change history, now filed as #105.

specs/competitive-positioning.md
jimador: Filed as #105. One append-only change record carries transaction time, so it closes both this and the old-and-new-value gap below.

specs/competitive-positioning.md
jimador: The remaining piece is filed as #105, an append-only change record with old and new value per field, attributed to the run or actor already carried by #95 and #101.

PR #83, opened 2026-08-30
feat(metamodel): schema versioning core with per-type governance

jimador:
PR 1 of the metamodel train (refs #45). `dice-metamodel` is a new pure-JVM module holding the schema-versioning contracts: `MetamodelVersion`, a content-addressed stamp of a declared schema whose fingerprint covers entity types, labels, full property signatures and relationships, canonically ordered and length-prefixed, with a derived, unforgeable `contentHash`; `GovernedTypeSelector` for per-type opt-in — you version what you declare, and ungoverned exploratory types never perturb a governed stamp; `DeclaredSchema`/`DeclaredSchemaSource`; and the `MetamodelVersionStore` contract. Declared renames ride `SchemaAliases` (**EXPERIMENTAL**): a declaration states the names a type or property used to go by, so a later diff pairs a rename and stops reading it as remove+add. Aliases hash only when non-empty, so every pre-alias `contentHash` is byte-stable, pinned by golden-digest tests.

- Schema attribution is answered through run lineage. The `DiceMetadataKeys.METAMODEL_VERSION` metadata key is removed; the extraction run carries the declared schema's `contentHash` and per-proposition attribution resolves through `PRODUCED_BY_RUN`. `docs/design/metamodel-versioning.md` states the mechanism and that the coordinator may write a denormalised copy once it exists.
- The doc now states context scope: versions carry no context dimension by design — one declared schema across all tenants, while drift reports and observation are per `ContextId`.
- `StampProvenance` stayed out: nothing constructs a value outside deserialization, so the type returns with the first stamping caller that records it.
- Naming and doc passes across the module.

**Breaking changes: none on any released surface.** The module is new; `@JvmOverloads` and separate overloads keep every previously shipped JVM descriptor byte-identical. The removed metadata key existed only inside this stack. The changed Kotlin data-class synthetics on `PropertySignature` are the stated boundary: Kotlin callers recompile.

**Opt-in:** declaring a `DeclaredSchemaSource` bean is the trigger for everything downstream; with no declared schema, nothing in this train activates.

**The governance model at a glance.** A type reaches the stamp only by being declared and selected:

```mermaid
flowchart LR
    A[DeclaredSchemaSource bean] --> C[DeclaredSchema]
    B[GovernedTypeSelector] --> C
    C --> D[MetamodelVersion]
    D --> E[entityTypeNames]
    D --> F[entityTypeLabels, full closure]
    D --> G[property signatures: name, kind, type, cardinality]
    D --> H[entityTypeAliases, former names]
    D --> I[relationships, attributed to declaring type]
    E --> J[contentHash: derived, unforgeable]
    F --> J
    G --> J
    H --> J
    I --> J
    K[ungoverned exploratory types] -. excluded .-> J
```

**Stack:** #84 (Drivine version store) and the later slices are stacked on this branch.


dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,570 @@
...
+ * write starts failing at the driver. A storage backend sizing a column has to size it in bytes: a
+ * 256-character value can reach 1024 bytes in UTF-8, and more once surrogate pairs are involved, so
+ * the column needs headroom rather than a matching 256.
+ *
+ * Experimental: shape may change before 1.0.
+ *
+ * @property actor Who asked for the stamp. Null when the host didn't say.
+ * @property trigger What prompted it. Null when the host didn't say.
+ */
+data class StampProvenance @JvmOverloads constructor(
```
jimador: Can we leave stamp provenance out until something actually produces it? Nothing in this stack constructs a non-null value outside deserialization, so this commits constructor, storage, and length semantics ahead of a caller.

james-gsb:
Addressed the review comment: StampProvenance is out of this PR entirely (fbbecf3). Verified first across the whole stack that nothing constructs a non-null value outside row-mapper deserialization — the one stamping call site never sets a cause, so the coalesce rules only ever moved nulls. The removal is hash-neutral (the pinned golden digest is byte-identical) and the design doc records the re-adoption trigger: provenance returns with the first stamping caller that records it. #84 drops the matching persistence in the same push.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Confirmed and removed. The only construction site on the whole stack was the row mapper reading the stored property back, so `StampProvenance` and its plumbing are gone from this branch — versions carry no provenance field until something actually produces one.

jimador:
The stack never writes the metamodel stamp that this design says a proposition carries. `docs/design/metamodel-versioning.md:51-53` names `dice.metamodel.version`, but the only production occurrence is the constant declaration in `DiceMetadataKeys`; no extraction or persistence path sets it, including the completed #83-#88 stack.

That leaves the version store unable to answer which schema a proposition was extracted under, and later drift selection cannot distinguish knowledge produced under the current declaration from older knowledge. Either the real extraction path needs to stamp canonical propositions with the declared content hash, or this PR needs to narrow the contract and name the later slice that owns it. The current complete stack does neither.

jimador:
Follow-up to the stamp comment above, because the other open stack changes the answer. The extraction-run stack carries `ExtractionRun.fingerprints.metamodelFingerprint` (#95, `ExtractionRunEnvelope.kt:41`) and it has no producer either. Two attribution mechanisms for one fact and nothing writes either. Pick the run. The coordinator on that stack resolves the `DeclaredSchemaSource` stamp and writes `contentHash` into the run, and a proposition's schema version is reachable through `PRODUCED_BY_RUN`. This PR then narrows `docs/design/metamodel-versioning.md:51-53` to say that, and `DiceMetadataKeys.METAMODEL_VERSION` either goes or is documented as a denormalised copy the coordinator writes. Full comment on #101.

One scope question for the doc. `MetamodelVersion` and `MetamodelVersionStore` have no context dimension, while drift reports and observation are per `ContextId`. One declared schema across all tenants is the right default for `me`, where the `DataDictionary` is application-wide. Say so in the doc, because a reader coming from the drift store will assume versions are scoped too.

jimador:
Done, on this branch as "Answer schema attribution through run lineage".

**One mechanism: the run.** `DiceMetadataKeys.METAMODEL_VERSION` is removed, together with its stamping test. `docs/design/metamodel-versioning.md:51-53` now says what you asked for: the extraction run carries the declared schema's `contentHash`, resolved by the coordinator from the host's `DeclaredSchemaSource`, and a proposition's schema version is reachable through `PRODUCED_BY_RUN`. The doc also states that the coordinator may write a denormalised per-proposition copy once it exists. The CHANGELOG records the key's removal as breaking for any holder of the constant; none exists outside this stack.

This corrects my earlier stamping-contract reply above. That reply described a defined metadata key waiting for a producer. The key is gone. Attribution answers through run lineage, and the coordinator — the run's first real writer — is recorded for the next wave, so until it lands the mechanism is defined and unwritten, which the doc says plainly.

**Context scope.** Stated in the same commit: the `MetamodelVersionStore` section now says versions carry no context dimension by design. One declared schema covers all tenants — the right default where the `DataDictionary` is application-wide — while drift reports and observation stay per `ContextId`. A reader arriving from the drift store now finds that spelled out.

jimador:
One leftover from the attribution rework. `CHANGELOG.md:44` records `DiceMetadataKeys.METAMODEL_VERSION` as removed and breaking, and `CHANGELOG.md:48` immediately after records it as added and additive. The key exists in neither the old nor the new tip. Drop the second entry.

james-gsb:
Dropped. The additive entry was already absent on this branch; it survived one level up on #84 through a bad changelog merge resolution during the restack. Removed there: `Drop the superseded stamping-contract changelog entry` (14780fc). The stack top now carries a single `METAMODEL_VERSION` entry, the breaking one — `grep -c "stamping contract"` at the top returns 0.

This round also marks the versioning surface: `@ApiStatus.Experimental` on `MetamodelVersion`, `PropertySignature`, `SchemaAliases`, `GovernedTypeSelector`, `DeclaredSchema`, `DeclaredSchemaSource` and `MetamodelVersionStore` (6c04af2).

jimador:
Dropped. The additive entry was already absent on this branch; it survived one level up on #84 through a bad changelog merge resolution during the restack. Removed there: `Drop the superseded stamping-contract changelog entry` (14780fc). The stack top now carries a single `METAMODEL_VERSION` entry, the breaking one — `grep -c "stamping contract"` at the top returns 0.

This round also marks the versioning surface: `@ApiStatus.Experimental` on `MetamodelVersion`, `PropertySignature`, `SchemaAliases`, `GovernedTypeSelector`, `DeclaredSchema`, `DeclaredSchemaSource` and `MetamodelVersionStore` (6c04af2).

docs/design/metamodel-versioning.md
```diff
@@ -0,0 +1,354 @@
...
+so the hash a piece of extracted data carries can be resolved back into the schema shape it stood
+for.
+
+## Why a version is a content hash
+
+The live schema is mutable, so stamping means taking an immutable snapshot of it.
+`MetamodelVersion.from(dataDictionary)` snapshots sorted entity type names, the full label set per
+type, the full property *signature* set per type, and sorted relationship descriptors, then
+fingerprints them as a SHA-256 `contentHash`. A proposition's schema attribution is answered through
+the run that produced it (PRODUCED_BY_RUN); the run record carries the declared schema's content hash,
```
jimador, line 50: The stack never writes the metamodel stamp that this design says a proposition carries. The design names `dice.metamodel.version`, but the only production occurrence is the constant declaration in `DiceMetadataKeys`; no extraction or persistence path sets it, including the completed #83-#88 stack.

That leaves the version store unable to answer which schema a proposition was extracted under, and later drift selection cannot distinguish knowledge produced under the current declaration from older knowledge. Either the real extraction path needs to stamp canonical propositions with the declared content hash, or this PR needs to narrow the contract and name the later slice that owns it.

Follow-up, because the other open stack changes the answer. The extraction-run stack carries `ExtractionRun.fingerprints.metamodelFingerprint` (#95, `ExtractionRunEnvelope.kt`) and it has no producer either. Two attribution mechanisms for one fact and nothing writes either. Pick the run. The coordinator on that stack resolves the `DeclaredSchemaSource` stamp and writes `contentHash` into the run, and a proposition's schema version is reachable through `PRODUCED_BY_RUN`. This PR then narrows this section to say that, and `DiceMetadataKeys.METAMODEL_VERSION` either goes or is documented as a denormalised copy the coordinator writes.

jimador: Done, on this branch as "Answer schema attribution through run lineage".

**One mechanism: the run.** `DiceMetadataKeys.METAMODEL_VERSION` is removed, together with its stamping test. This section now says what you asked for: the extraction run carries the declared schema's `contentHash`, resolved by the coordinator from the host's `DeclaredSchemaSource`, and a proposition's schema version is reachable through `PRODUCED_BY_RUN`. The doc also states that the coordinator may write a denormalised per-proposition copy once it exists. The CHANGELOG records the key's removal as breaking for any holder of the constant; none exists outside this stack.

The key is gone. Attribution answers through run lineage, and the coordinator, the run's first real writer, is recorded for the next wave, so until it lands the mechanism is defined and unwritten, which the doc says plainly.

docs/design/metamodel-versioning.md
```diff
@@ -0,0 +1,354 @@
...
+cause would give one schema as many identities as it had causes.
+
+## History accumulates
+
+`MetamodelVersionStore` is a port with four operations: `saveVersion`, `latestVersion`,
+`versionHistory`, and `findVersion(schemaName, contentHash)`. There is no delete. Correlating what
+a graph holds today against the stamp it was extracted under only works while the old stamps are
+still there.
+
+`MetamodelVersion` and `MetamodelVersionStore` carry no context dimension — one declared schema
```
jimador, line 315: One scope question for the doc. `MetamodelVersion` and `MetamodelVersionStore` have no context dimension, while drift reports and observation are per `ContextId`. One declared schema across all tenants is the right default for a host whose `DataDictionary` is application-wide. Say so here, because a reader coming from the drift store will assume versions are scoped too.

jimador: Stated in the same commit: this section now says versions carry no context dimension by design. One declared schema covers all tenants, the right default where the `DataDictionary` is application-wide, while drift reports and observation stay per `ContextId`. A reader arriving from the drift store now finds that spelled out.

CHANGELOG.md
```diff
@@ -0,0 +1,47 @@
...
+  GovernedTypeSelector, int, Object)` with a wider descriptor and broken any
+  caller already compiled against it. `MetamodelJavaCompatTest` calls every
+  static form. The changed Kotlin synthetic constructor, `copy`, `copy$default`
+  and `componentN` signatures on `PropertySignature` are the accepted boundary:
+  Kotlin callers recompile, and no consumer holds a compiled reference to them.
+
+- Schema attribution mechanism: per-proposition version is answered through the
+  run that produced the proposition (PRODUCED_BY_RUN). The run record carries
+  the declared schema's content hash, resolved by the extraction coordinator from
+  the host's DeclaredSchemaSource. The `DiceMetadataKeys.METAMODEL_VERSION`
```
jimador, line 44: One leftover from the attribution rework. This entry records `DiceMetadataKeys.METAMODEL_VERSION` as removed and breaking, and an entry immediately after records it as added and additive. The key exists in neither the old nor the new tip. Drop the second entry.

jimador: Dropped. The additive entry was already absent on this branch; it survived one level up on #84 through a bad changelog merge resolution during the restack. Removed there: `Drop the superseded stamping-contract changelog entry` (14780fc). The stack top now carries a single `METAMODEL_VERSION` entry, the breaking one.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+ * @property aliases Names this property used to go by, so a rename pairs up instead of reading as a
+ *   removal and an addition. Declared through [SchemaAliases] and empty unless someone declared
+ *   them. Part of the signature, so an alias-only edit moves [MetamodelVersion.contentHash].
+ *   Held as handed in, including a mutable set: a `data class` component can't be normalised on the
+ *   way through, since a constructor `val` takes no initialiser and the generated `copy` calls that
+ *   same constructor. [MetamodelVersion] therefore copies the set into an immutable one when it
+ *   takes a signature, and a signature that never reaches a stamp keeps whatever set it was built
+ *   with. Experimental: shape may change before 1.0.
+ */
+@ApiStatus.Experimental
```
jimador, line 52: Nothing in this stack carried `@ApiStatus.Experimental` while the CHANGELOG called these surfaces experimental in prose. Mark the public types so a consumer sees it in the IDE.

jimador: Marked on this branch (6c04af2): `MetamodelVersion`, `PropertySignature`, `SchemaAliases`, `GovernedTypeSelector`, `DeclaredSchema`, `DeclaredSchemaSource` and `MetamodelVersionStore`.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DeclaredSchemaSource.kt
```diff
@@ -0,0 +1,128 @@
...
+ * @property relationshipTypeNames The bare relationship type names [version] allows.
+ */
+@ApiStatus.Experimental
+class DeclaredSchema(
+    val version: MetamodelVersion,
+    relationshipTypeNames: Set<String>,
+) {
+
+    /** Copied into a JVM-immutable set so the declaration can't drift from its stamped [version]. */
+    val relationshipTypeNames: Set<String> = java.util.Set.copyOf(relationshipTypeNames)
```
igordayen, line 42: kotlin setOf?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DeclaredSchemaSource.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    override fun equals(other: Any?): Boolean =
```
igordayen: kotlin is supposed to auto-generate equals / hashcode

dice-metamodel/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+        </dependency>
+
+        <!--
+            https://mvnrepository.com/artifact/org.jetbrains/annotations
+            For Experimental annotation
+        -->
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>26.0.2</version>
```
igordayen: please consider meta property.  Version - Not inherited from embabel-parent-dependencies?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/GovernedTypeSelector.kt
```diff
@@ -0,0 +1,59 @@
...
+ * val governed = setOf("Person", "Company")
+ * val version = MetamodelVersion.from(dataDictionary, GovernedTypeSelector { it.name in governed })
+ * ```
+ *
+ * Selecting a subset changes which types the stamp covers, and leaves the encoding alone. Adding an
+ * ungoverned type to the dictionary leaves the content hash as it was, while touching a governed
+ * one changes it.
+ */
+@ApiStatus.Experimental
+fun interface GovernedTypeSelector {
```
igordayen, line 42: what about alternative chained syntax:

MetamodelVersion.from(dataDictionary).governedBy(governed)?

may potentially involve less maintenance.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+        /**
+         * Total order used to canonicalise a property set before hashing. Aliases come last, so a
+         * set holding two signatures that differ only in their aliases still sorts the same way
+         * every time.
+         */
+        private val ORDER = compareBy<PropertySignature>({ it.name }, { it.kind }, { it.type }, { it.cardinality })
+            .thenComparator { left, right -> compareAliases(left.aliases, right.aliases) }
+
+        /** Compare two alias sets as sorted lists: element by element, then by size. */
+        private fun compareAliases(left: Set<String>, right: Set<String>): Int {
```
igordayen, line 89: consider more readable Kotlin:

```
private fun compareAliases(left: Set<String>, right: Set<String>): Int {
    val leftIt = left.sorted().iterator()
    val rightIt = right.sorted().iterator()

    while (leftIt.hasNext() && rightIt.hasNext()) {
        val comparison = leftIt.next().compareTo(rightIt.next())
        if (comparison != 0) return comparison
    }

    // If one sequence runs out, the shorter one comes first
    return leftIt.hasNext().compareTo(rightIt.hasNext())
}

```

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        /**
+         * Read the structural signature off a property definition.
+         *
+         * @param property The property definition to summarise.
+         * @return Its structural signature.
+         */
+        @JvmStatic
+        fun of(property: PropertyDefinition): PropertySignature = when (property) {
```
igordayen, line 104: so Kind the translation of PropertyDefinition subclass?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+
+    init {
+        // Only types named in entityTypeNames are walked when hashing, so a map entry keyed by
+        // anything else never reaches contentHash, and two stamps holding different labels or
+        // properties could share a hash and the store's natural key with it. `this.` is required
+        // here: the constructor parameters are still in scope and would shadow the copied fields.
+        val known = this.entityTypeNames.toSet()
+        val strayLabelKeys = this.entityTypeLabels.keys - known
+        require(strayLabelKeys.isEmpty()) {
+            "entityTypeLabels is keyed by types missing from entityTypeNames: ${strayLabelKeys.sorted()}. " +
```
igordayen, line 187: please avoid concats, consider multi-line string with backticks

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+            relationshipNames.forEach { appendSized(it) }
+        }
+
+        val digest = MessageDigest.getInstance("SHA-256")
+        val hashBytes = digest.digest(hashInput.toByteArray(Charsets.UTF_8))
+        return hashBytes.joinToString("") { "%02x".format(it) }
+    }
+
+    /** Structural equality: same schema name, types, labels, property signatures, relationships, and aliases. */
+    override fun equals(other: Any?): Boolean {
```
igordayen, line 279: can be constructed by  utility  EquilsBuilder?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+         * Two same-named domain types can each declare `age` with a different shape, and the union
+         * keeps both signatures. A comparison has no way to say which of the two an old name refers
+         * to, so the declaration is refused and the name compares as a removal plus an addition.
+         */
+        private fun requireNoAliasesOnDuplicateNames(
+            entityTypeProperties: Map<String, Set<PropertySignature>>,
+        ) {
+            entityTypeProperties.forEach { (typeName, signatures) ->
+                signatures
+                    .groupBy { it.name }
```
igordayen, line 391: too cryptic to read without liny-by-line comments

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+                    .filterValues { withName -> withName.size > 1 }
+                    .forEach { (propertyName, withName) ->
+                        val declared = withName.flatMap { it.aliases }.distinct().sorted()
+                        require(declared.isEmpty()) {
+                            "Property '$propertyName' on entity type '$typeName' has " +
```
igordayen, line 396: same as above and everywhere about multi-line comments

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+ *   through [SchemaAliases] and empty unless someone declared them. Hashed, so an alias-only edit
+ *   moves [contentHash]. Experimental: shape may change before 1.0.
+ * @property contentHash SHA-256 hex digest of the schema's entity types, label sets, property
+ *   signatures, type aliases, and allowed relationships. The schema name is excluded so that two
+ *   structurally identical schemas are equal regardless of how they are named. Stable across JVM
+ *   restarts. Any structural change produces a different hash, including a property's type changing
+ *   on a type whose name is unchanged.
+ */
+@ApiStatus.Experimental
+class MetamodelVersion @JvmOverloads constructor(
```
igordayen, line 157: is Version in name fully justified?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
```diff
@@ -0,0 +1,517 @@
...
+         *   governs all of them.
+         * @param aliases Former names for the schema's types and properties. [SchemaAliases.NONE]
+         *   declares none. Experimental: shape may change before 1.0.
+         * @return An immutable version stamp covering the governed subset.
+         * @throws IllegalArgumentException when a declared type name appears in another type's
+         *   alias set, or when aliases are declared for a property name the governed types hold
+         *   more than one signature for.
+         */
+        @JvmStatic
+        fun from(
```
igordayen, line 467: from(Dictionary).governedBy(...).withAliaes - more fluent API / DSL style

more in spirit of embabel-agent-api

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersionStore.kt
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * The default scans [versionHistory], which is correct for any implementation but reads the
+     * whole history to answer a keyed question. A backend that can push the lookup down to the
+     * database (a keyed `MATCH` rather than an in-memory `filter`) should override it.
+     *
+     * @param schemaName The schema the version belongs to.
+     * @param contentHash The [MetamodelVersion.contentHash] to find.
+     * @return The matching version, or `null` if this schema has no version with that hash.
+     */
+    fun findVersion(schemaName: String, contentHash: String): MetamodelVersion? =
```
igordayen, line 92: contentHash is derived from MetaModelVersion instance, is my understanding accurate?

dice-metamodel/src/test/java/com/embabel/dice/metamodel/MetamodelJavaCompatTest.java
```diff
@@ -0,0 +1,191 @@
...
+    private static DomainType type(String name) {
+        return new DynamicType(name, "", List.of(), List.of(), true);
+    }
+
+    private static DataDictionary goldenSchema() {
+        return DataDictionary.fromDomainTypes("golden-schema", List.of(type("Person"), type("Company")));
+    }
+
+    @Test
+    void theFourArgumentPropertySignatureConstructorStillExists() {
```
igordayen, line 56: for unit test we usually use natural description as english sentence:
  ` the four-argument property signature etc...`

igordayen (commented):
@jimador - first batch of  manual review comments, thank you

igordayen:
from claude - 1st batch:

```
──────────────────────────────────────┬──────────────────────────────────────────────────────────────┐
  │               Comment                │                            Valid?                            │
  ├──────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ compareAliases readability rewrite   │ Yes — the indexed loop is more readable than index-based     │
  │                                      │ access                                                       │
  ├────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ Avoid string concat → multiline         │ Yes — valid Kotlin style                                  │
  │ template                                │                                                           │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ Fluent API from(dict).governedBy(...)   │ Debatable — adds intermediate builder objects for         │
  │                                         │ marginal readability gain                                 │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ Test method names as English sentences  │ Yes — standard Kotlin/JUnit5 style                        │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ is Version in name fully justified?     │ Reasonable question — MetamodelStamp or MetamodelSnapshot │
  │                                         │  might read better                                        │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ contentHash is derived from instance?   │ Valid clarification question — answer is yes              │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ too cryptic without comments (lines     │ Valid — hash computation code with no inline explanation  │
  │ 369/374)                                │                                                           │
  ├─────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ Version not in                          │ Valid housekeeping question                               │
  │ embabel-parent-dependencies             │                                                           │
  └─────────────────────────────────────────┴───────────────────────────────────────────────────────────┘

```

igordayen:
feedback from Claude:

Overall: Well-engineered. The design decisions are sound and thoroughly documented. A few issues worth
  raising:

  ---
  1. equals / hasSameContentAs asymmetry — potential API trap
  
  MetamodelVersion.equals includes schemaName. contentHash and hasSameContentAs exclude it by design. This
  creates two notions of equality on the same object. The store's natural key is (schemaName, contentHash),
  so equality-based data structures (e.g. HashSet<MetamodelVersion>) behave differently from a store lookup
  by contentHash. Nothing is wrong, but a reader arriving from the store will be confused. equals KDoc
  should explicitly say it includes schemaName while hasSameContentAs does not, with a cross-reference.

  ---
  2. Validation runs twice in from() — code duplication risk

  requireNoTypeAliasReuse and requireNoAliasesOnDuplicateNames are called in both the from() factory and the
  init block. The comment explains the attribution reason (fail at the call site). That's valid. But if the
  validation logic ever changes, the two call sites can silently diverge. Consider a single validate(names,
  aliases, properties) function called from both — so there's one body, two call sites.

  ---
  3. MetamodelVersionStore.findVersion default scans full history
  
  The interface KDoc acknowledges this: "reads the whole history to answer a keyed question." But there's no
  signal to implementors about the performance expectation. If history grows large (which it will in a
  long-lived schema), the default is quietly O(n). Either mark it @Deprecated to push implementors to
  override, or add a stronger advisory in the KDoc noting this is unsuitable for production use without
  override.

  ---
  4. SchemaAliases has manual equals/hashCode/toString — same pattern as DeclaredSchema

  The same reason applies: java.util.Set.copyOf() in the constructor body means a data class wouldn't
  include typeAliases / propertyAliases as components. Consistent with the rest, but worth noting it's a
  deliberate pattern across the module, not an oversight. A module-level design note would prevent reviewers
  raising this on every class.

  ---
  5. from() builds entityTypeProperties in a loop that's already a groupBy — two traversals

  val entityTypeProperties = governedTypes
      .groupBy { it.name }
      .mapValues { (typeName, types) ->
          types.flatMap { type -> type.properties.map { ... } }.toSet()
      }

  entityTypeLabels does a separate .groupBy { it.name }. Both traverse governedTypes and group by the same
  key. Could be a single pass: groupBy once, then derive both labels and properties from the grouped map.
  Minor, but this is a hot path if called frequently.

  ---
  6. toString() uses + concatenation across multiple lines

  MetamodelVersion.toString() concatenates 7 fields with +. In Kotlin this should be a multiline string
  template or buildString. Minor style issue but inconsistent with the otherwise idiomatic codebase.

  ---
  7. No concern about contentHash stability on JVM charset differences
  
  The hash is computed as hashInput.toByteArray(Charsets.UTF_8) — explicitly UTF-8. Good. The KDoc says
  "Stable across JVM restarts" which is true given deterministic ordering and UTF-8. The golden-digest test
  pins this. No issue — just confirming this is correctly handled.

  ---
  Summary: The core correctness invariants (length-prefixed hash encoding, immutable collections,
  deduplication in init) are solid. Items 1 and 3 are the most actionable.

igordayen:
i just realized that hashcode includes already schema name, therefore asked claude to validate:

why search by (name, hashcode), is not hashcode sufficient?                                          

● Good catch. If contentHash already excludes schemaName and is a SHA-256 of the full structural content,
  then contentHash alone is the natural key — it's globally unique by construction. The (schemaName, 
  contentHash) composite key in MetamodelVersionStore is redundant.

  The only scenario where schemaName adds value as part of the key is if you want to scope history queries
  per schema (latestVersion(schemaName), versionHistory(schemaName)). But that's a query filter, not a
  uniqueness requirement — contentHash alone already guarantees uniqueness.

  Worth raising: either drop schemaName from the store's natural key and keep it only as a query dimension,
  or document explicitly why it's in the key. Right now the KDoc says (schemaName, contentHash) is the
  natural key without justifying why contentHash alone isn't sufficient.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DeclaredSchemaSource.kt
jimador: Deliberate. Kotlin's `toSet()` returns a `LinkedHashSet` behind the read-only `Set` type, so a caller can cast it back to `MutableSet` and mutate a stamped declaration's field. `java.util.Set.copyOf` returns a genuinely immutable set that throws on any mutation, which is what the line above it promises. Order is not load-bearing here: these names are read for membership and equality only.

`MetamodelVersion` does the same thing through its own `immutableCopy`, so the two agree.

dice-metamodel/pom.xml
jimador: Fixed. Neither `embabel-build-parent` nor the `embabel-agent-dependencies` BOM manages `org.jetbrains:annotations`, so each module was pinning its own. That had already drifted: `dice` pinned 26.0.1 and `dice-metamodel` 26.0.2.

There is now a `jetbrains.annotations.version` property with a `dependencyManagement` entry in `dice-parent`, and both modules inherit it.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Taken. It now zips the two sorted lists and keeps the size comparison explicit at the end, so the shorter set still sorts first without the boolean comparison doing that work.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Yes. `Kind` is the structural category read off the `PropertyDefinition` subclass: `DomainTypePropertyDefinition` is `REFERENCE`, `NamedPropertyDefinition` is `VALUE`, anything else is `UNKNOWN`.

The stamp keeps the category and the type name, never the definition, so a new subclass upstream lands as `UNKNOWN` and leaves every existing content hash where it was.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Keeping the concatenation here. The message is one sentence, so a raw string would put a newline in the middle of it and split the line an operator greps for in a log.

`embabel-agent` builds its long messages the same way, for example `ActionMethodArgumentResolver.kt:198` and `AgentMetadataReader.kt:775`.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: That would mean depending on commons-lang3, which dice does not, and the reflective form walks the fields on every call. This type is compared and hashed on the store's natural-key path.

`hashCode` already delegates to `java.util.Objects.hash`. `equals` stays explicit because the class is not a data class: every field is defensively copied in the body, so the generated equality would read the constructor parameters.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Reading it in words: for each governed type, group that type's property signatures by property name, keep only the names holding more than one signature, and refuse the declaration when any of those names also declares an alias.

The refusal is the point. Two same-named types can each declare `age` with a different shape and the union keeps both signatures, so an old name pointing at that property cannot say which of the two it meant.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Same answer as above: the message stays one line so it greps as one line, matching how `embabel-agent` builds its own.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: It is a version in the sense of Hibernate's `@Version`: an identity you compare, not a number that increases. The stamp is a content hash plus the identity it covers, and the store reads as `latestVersion` and `versionHistory` off it.

If you prefer stamp or fingerprint, I will take the rename, but it touches six stacked PRs, so it wants to be one pass and not a change here.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersionStore.kt
jimador: Accurate. It is SHA-256 over the stamp's structural fields: entity type names, label sets, property signatures, type aliases and relationship names. The schema name is excluded, so two structurally identical schemas hash the same however they are named.

The store treats `(schemaName, contentHash)` as the natural key, and the Drivine row mapper recomputes the hash on read and rejects the row when it disagrees with the stored one.

dice-metamodel/src/test/java/com/embabel/dice/metamodel/MetamodelJavaCompatTest.java
jimador: Done, all twelve now carry a `@DisplayName` sentence. Java method names cannot hold spaces, so that is the JUnit 5 equivalent of the backtick names the Kotlin tests use.

igordayen (approved):

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/GovernedTypeSelector.kt
jimador: I really like this idea. I will add a PR to the stack that addresses this change.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: I really like this idea. I will add a PR to the stack that addresses this change.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/GovernedTypeSelector.kt
jimador: Agreed, and built as #106 at the top of the stack.

`MetamodelVersion.stamping(dictionary).governedBy(governed).stamp()`, with `governedBy` taking either a selector or a set of type names. It could not hang off `from(dictionary)` itself, since that already returns a stamp and changing its return type breaks every compiled caller, so the chain starts at a new factory and delegates to the existing overloads.

The example on this KDoc now shows both forms.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersion.kt
jimador: Done in #106, close to what you wrote: `MetamodelVersion.stamping(dictionary).governedBy(governed).withAliases(aliases).stamp()`.

`MetamodelStamping` is an immutable data class carrying the three inputs, so every step copies and a partly built stamping can be held in a field and finished more than once. It finishes as either a stamp or a declaration, which is what keeps the two halves of a declaration from disagreeing about what is governed.

The chain is identical from Java, since each step is a plain method taking one argument. Four tests in `MetamodelJavaCompatTest` cover the chain, the declaration terminal, the copy-on-step behaviour and the constructor arities `@JvmOverloads` generates.

These three `from` overloads stay as the short forms.

jimador:
Went through all three. The table in the first batch is covered by the thread replies. On the second batch and the key question:

**1. `equals` and `hasSameContentAs`.** Real. Both KDocs now cross-reference: `equals` compares the schema name, `hasSameContentAs` and `contentHash` leave it out, so a `Set<MetamodelVersion>` keys the way the store does and `hasSameContentAs` asks whether two schemas have the same shape.

**2. Validation in two places.** Real. The two refusals now run through one `requireDeclarable`, called from `from` and from the constructor. One body, two call sites, and they cannot drift on which checks apply.

**3. `findVersion` default.** The Drivine store already overrides it with a keyed `MATCH` (#84). The KDoc now says the default is for the in-memory reference and test doubles, and that a durable backend should override it because a schema's history only grows. Not deprecating it: it is the correct contract and the reference implementation.

**4. Manual `equals` and `hashCode`.** Added a section to `docs/design/metamodel-versioning.md`, "Plain classes, not data classes", so the pattern reads as one module decision. The reason is that each class copies its arguments into an immutable collection in the body, which a data class cannot express.

**5. Two traversals in `from`.** Real. It groups by name once now and derives labels, properties and the governed name set from that one grouping.

**6. `toString` concatenation.** Keeping it, for the same reason as the thread replies: one-line messages stay one line, and `embabel-agent` builds its own the same way.

**7. UTF-8.** Agreed, and the golden-digest test is what pins it.

**Is the schema name redundant in the key?** No, and the store KDoc now says why. `contentHash` excludes the name on purpose, so two schemas with the same shape share a hash: a schema and its staging copy, or a schema forked under a new name. History is per schema, which is what `latestVersion` and `versionHistory` answer, so the same content has to be a separate record under each name. Keyed on the hash alone, one schema adopting a shape another had earlier would land on the other schema's record and pull that history into its own. The name is what keeps two schemas' histories apart.

PR #85, opened 2026-08-30
feat(metamodel): observed-schema and diff contracts

jimador:
PR 3 of the metamodel train, stacked on #84. This slice defines the two comparisons the governance loop runs on. `ObservedSchema`/`ObservedSchemaSource` is the SPI a storage backend implements to report what a live graph actually contains — names only, since a graph exposes labels and relationship types without signatures. `MetamodelDiff`/`MetamodelChange` is a sealed change taxonomy: a property that kept its name and changed type, cardinality or kind reports as one `PropertySignatureChanged` entry. `MetamodelDiffer` compares two declared versions; `DeclaredObservedDiffer` compares a declaration against the graph. Declared renames pair in the diff (**EXPERIMENTAL**): `EntityTypeRenamed`, `PropertyRenamed` and `EntityTypeAliasesChanged` are new sealed members, pairing is one-to-one on the residual sets after name matching, comparison substitutes old-for-new in reference targets and label sets only, and a pure rename yields one change. Accumulated former names join the declared side of the observed diff, so old-labeled data under a declared rename is never drift.

- Declared and observed types compare by own label. `JvmType.name` is a fully qualified class name while the graph reports simple labels, so a healthy FQN schema previously came back both unobserved and drifted at once. `DeclaredSchema.entityTypeOwnLabels` (via `DeclaredSchema.ownLabelOf`) fixes both directions; reported spellings stay as the stamp declares them.
- Known-but-ungoverned types stay out of observed drift, covering their own labels too.
- Contested rename claims report as such; the differ never picks a winner silently.
- `TypeIdentity` added (**EXPERIMENTAL**): the contract for mapping external type spellings onto declared names, with no implementation and no wiring yet.

**Breaking changes:** one stated source-level exception — `MetamodelChange` is sealed and gains members, so an external exhaustive `when` over it needs new branches. Binary compatibility is untouched; consumers recompile. Everything else is additive.

**The two comparisons.** Declared-vs-declared answers what a schema edit did; declared-vs-observed answers what the live graph holds that the declaration does not cover:

```mermaid
flowchart TD
    A[DeclaredSchema, version n] --> B[MetamodelDiffer.diff]
    C[DeclaredSchema, version n-1] --> B
    B --> D[declared vs declared: renames paired, signatures compared modulo renames]
    A --> E[DeclaredObservedDiffer.diffAgainstObserved]
    F[ObservedSchema, names only from a live graph] --> E
    E --> G[declared vs observed: undeclared types and relationship names, judged by own label]
    D --> H[MetamodelDiff: sealed MetamodelChange taxonomy]
    G --> H
```

**Next in stack:** #86 drift checking and quarantine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,18 +77,18 @@ and the consumer PRs that deliver it).
   property that existed before keeps its name, meaning, and encoding, and the two
   new alias fields are absent when nothing declares them.
 - Schema diffing contracts in `dice-metamodel`: `MetamodelDiffer` compares two declared
-  stamps and returns a `MetamodelDiff` — an ordered, canonically sorted list of sealed
-  `MetamodelChange` entries. The taxonomy covers property *signatures*, not just names:
-  alongside `EntityTypeAdded`/`Removed`/`Modified` and `RelationshipAdded`/`Removed`, a
+  stamps and returns a `MetamodelDiff`, an ordered, canonically sorted list of sealed
+  `MetamodelChange` entries. The taxonomy covers property *signatures* as well as names.
+  Alongside `EntityTypeAdded`/`Removed`/`Modified` and `RelationshipAdded`/`Removed`, a
   `PropertySignatureChanged` pairs `before` and `after` for a property that kept its name
-  but changed value type, cardinality, or value-vs-reference kind — the case a name-only
-  diff reports as nothing at all. `DeclaredObservedDiffer` answers the other question,
-  comparing a `DeclaredSchema` against an `ObservedSchema` snapshot of a live graph and
-  separating drift (observed but undeclared, actionable) from unobserved (declared but
-  empty, normal). That comparison is names-only on the observed side, deliberately: a graph
-  reports labels and relationship types, never declared property shapes.
-  `ObservedSchemaSource` is the SPI a storage backend implements later; `ObservedSchema` is
-  a plain value type, so tests drive the whole comparison from a canned snapshot.
-  `StructuralMetamodelDiffer` implements both interfaces — deterministic, stateless, no
-  database and no LLM. No drift runner, no quarantine, no Spring, and no new dependency.
+  and changed value type, cardinality, or value-vs-reference kind, which a name-only diff
+  reports as no change at all. `DeclaredObservedDiffer` compares a `DeclaredSchema` against
+  an `ObservedSchema` snapshot of a live graph, separating drift (observed but undeclared,
+  actionable) from unobserved (declared but empty, normal). That comparison is names-only
+  on the observed side: a graph reports labels and relationship types, and cannot report
+  declared property shapes. `ObservedSchemaSource` is the SPI a storage backend implements
+  in a later slice; `ObservedSchema` is a plain value type, so tests drive the whole
+  comparison from a canned snapshot. `StructuralMetamodelDiffer` implements both
+  interfaces: deterministic, stateless, no database and no LLM. This slice adds no drift
+  runner, no quarantine, no Spring wiring, and no new dependency.
   **Compatibility: additive.** New types in an existing module; no existing API touched.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,3 +92,40 @@ and the consumer PRs that deliver it).
   interfaces: deterministic, stateless, no database and no LLM. This slice adds no drift
   runner, no quarantine, no Spring wiring, and no new dependency.
   **Compatibility: additive.** New types in an existing module; no existing API touched.
+
+- Declared renames in the diff, **EXPERIMENTAL** (shape may change before 1.0): three new
+  `MetamodelChange` members — `PropertyRenamed(typeName, before, after)`,
+  `EntityTypeRenamed(before, after)` and `EntityTypeAliasesChanged(typeName, before, after)`.
+  A rename declared through `SchemaAliases` now pairs instead of reading as a removal and an
+  addition. Pairing runs on what is left after the ordinary name matching, walks removed
+  names in sorted order against added entries in sorted order, and is one-to-one: one added
+  entry claiming two removed names takes the first and leaves the other an ordinary removal,
+  one removed name claimed by two added entries goes to the first and leaves the second an
+  ordinary addition, and a property name the type-merge path holds two signatures for falls
+  back to a removal and an addition. Paired properties are excluded from
+  `EntityTypeModified.addedProperties`/`removedProperties`, and an entry left empty by that
+  is not emitted. Type pairing matches the whole accumulated alias set, so a type renamed
+  twice still pairs across stamps that aren't adjacent, and it suppresses the
+  `EntityTypeAdded`/`EntityTypeRemoved` pair, reporting the type's other deltas under the new
+  name. After pairing, the older version is compared modulo the renames the diff found: old
+  name for new is substituted in `Kind.REFERENCE` signature targets and in label sets, and
+  nowhere else. A delta that vanishes under the substitution folds into `EntityTypeRenamed`,
+  and one that survives reports in substituted form (a referrer that moved `A → D` while `A`
+  was renamed to `B` reports `B → D`). `Kind.VALUE` type strings are left alone, so an entity
+  type named `Date` renaming to `Timestamp` does not rewrite every property declared as
+  holding a `Date` value. Rendered relationship descriptors are left alone too, so a
+  relationship touching a renamed endpoint still churns as a removal plus an addition; the
+  names inside a descriptor are free text and are never parsed. Alias-only edits have
+  representations, so an empty diff and an equal `contentHash` keep meaning the same thing: a
+  property whose aliases alone moved is an ordinary `PropertySignatureChanged`, and a type's
+  is an `EntityTypeAliasesChanged`. Declared former names join the declared side of
+  `diffAgainstObserved`, so data still carrying a renamed type's old label is not drift.
+  `MetamodelDiff.touchedEntityTypes` covers the new kinds, and contributes both names of a
+  rename.
+  **Compatibility: breaking for external exhaustive `when` expressions.** `MetamodelChange`
+  is a sealed interface, and three new members make any `when` over it outside this repo
+  non-exhaustive until it handles them. Stated and accepted: the taxonomy is designed to be
+  exhausted, and a consumer that silently treated a rename as an unhandled kind would treat
+  it as harmless. Nothing else changes for a schema that declares no aliases — pairing and
+  substitution are no-ops with an empty alias map, and the existing diff behavior, ordering
+  and output are unchanged. Binary compatibility is untouched; consumers recompile.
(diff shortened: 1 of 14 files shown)


dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/StructuralMetamodelDiffer.kt
```diff
@@ -0,0 +1,379 @@
...
+        added: Set<String>,
+        aliases: Map<String, Set<String>>,
+    ): Map<String, String> {
+        if (removed.isEmpty() || added.isEmpty() || aliases.isEmpty()) return emptyMap()
+
+        val candidates = added.sorted()
+        val claimed = mutableSetOf<String>()
+        val paired = LinkedHashMap<String, String>()
+        for (removedName in removed.sorted()) {
+            val match = candidates.firstOrNull { addedName ->
```
jimador: Two added types can claim the same former name here, and this picks the first alphabetically as the rename. Can we reject the ambiguity or report it instead of attaching the old type's changes to an arbitrary new type?

jimador: The differ no longer picks a winner. A contested claim in either direction — several added types naming one former name, or one new name claimed by several live former names — now surfaces as `AmbiguousEntityTypeRename` / `AmbiguousPropertyRename` changes with their own typed accessors, and none of the contested history attaches to any type.

jimador:
Partial governance is lost before the observed comparison. `DeclaredSchema.from` filters the dictionary to the governed types, then `StructuralMetamodelDiffer.diffAgainstObserved` compares every observed name against only that filtered stamp. A known exploratory type excluded by `GovernedTypeSelector` is therefore indistinguishable from an undeclared type; the runner turns it into a removal and the quarantine policy can stale propositions that were explicitly outside governance.

The declaration needs to retain the known-but-ungoverned type/relationship set for exclusion from observed drift. A selector governing no types must enforce no types. Please add the discriminating case: a dictionary with governed `Person` and ungoverned `Sighting`, an observation containing `Sighting`, and a live run that leaves the `Sighting` proposition active.

jimador:
Fixed in `747641a` (rebased through the stack).

`DeclaredSchema` now carries the known-but-ungoverned type and relationship names beside the governed stamp, as a third set distinct from the governed names and from declared aliases, and `diffAgainstObserved` excludes them from drift. A selector governing no types now enforces no types: every dictionary type stays ungoverned, and only a name the dictionary never declared reads as drift.

Both discriminating cases are in: governed `Person` with ungoverned `Sighting` observed, asserting no drift; and the empty-selector case, which also pins that a genuinely undeclared type is still reported. Mutation-checked.

The end-to-end assertion you named — a live run leaving the `Sighting` proposition `ACTIVE` — needs `DefaultDriftCheckRunner`, which lives on #86. I am carrying it there rather than duplicating a runner harness on this branch.

Gate: `dice` 1189/0, `dice-metamodel` 209/0, stack top green.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DeclaredSchemaSource.kt
```diff
@@ -29,30 +31,96 @@ import org.jetbrains.annotations.ApiStatus
  * declared relationships against what a graph holds needs those bare names, and that comparison is
  * the next slice.
  *
+ * An ungoverned type is still a known one: the host's dictionary named it, and a
```
jimador, line 34: Partial governance is lost before the observed comparison. `DeclaredSchema.from` filters the dictionary to the governed types, then `StructuralMetamodelDiffer.diffAgainstObserved` compares every observed name against only that filtered stamp. A known exploratory type excluded by `GovernedTypeSelector` is therefore indistinguishable from an undeclared type; the runner turns it into a removal and the quarantine policy can stale propositions that were explicitly outside governance.

The declaration needs to retain the known-but-ungoverned type/relationship set for exclusion from observed drift. A selector governing no types must enforce no types. Please add the discriminating case: a dictionary with governed `Person` and ungoverned `Sighting`, an observation containing `Sighting`, and a live run that leaves the `Sighting` proposition active.

jimador: Fixed in `747641a` (rebased through the stack).

`DeclaredSchema` now carries the known-but-ungoverned type and relationship names beside the governed stamp, as a third set distinct from the governed names and from declared aliases, and `diffAgainstObserved` excludes them from drift. A selector governing no types now enforces no types: every dictionary type stays ungoverned, and only a name the dictionary never declared reads as drift.

Both discriminating cases are in: governed `Person` with ungoverned `Sighting` observed, asserting no drift; and the empty-selector case, which also pins that a genuinely undeclared type is still reported. Mutation-checked.

The end-to-end assertion you named, a live run leaving the `Sighting` proposition `ACTIVE`, needs `DefaultDriftCheckRunner`, which lives on #86. It is carried there.

Gate: `dice` 1189/0, `dice-metamodel` 209/0, stack top green.

CHANGELOG.md
```diff
@@ -76,3 +76,99 @@ and the consumer PRs that deliver it).
   module dependency; no existing API touched. Stored nodes stay readable: every
   property that existed before keeps its name, meaning, and encoding, and the two
   new alias fields are absent when nothing declares them.
+- Schema diffing contracts in `dice-metamodel`: `MetamodelDiffer` compares two declared
```
igordayen, line 79: would ike to keep in GitHub?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/StructuralMetamodelDiffer.kt
```diff
@@ -0,0 +1,542 @@
...
+ *
+ * A rename has to be unmistakable to pair. When two new types both declare `Person` as a former
+ * name, the declaration is saying two things at once, and the differ pairs neither: both read as
+ * ordinary additions, `Person` reads as an ordinary removal, and an
+ * [MetamodelChange.AmbiguousEntityTypeRename] says which claim was set aside. See
+ * `docs/design/metamodel-diff.md`.
+ *
+ * Stateless and thread-safe; one shared instance is fine.
+ */
+class StructuralMetamodelDiffer : MetamodelDiffer, DeclaredObservedDiffer {
```
igordayen, line 58: is it used outside DICE as a framework.  reason for ask - visibility scope as public. thanks

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/StructuralMetamodelDiffer.kt
```diff
@@ -0,0 +1,542 @@
...
+
+    override fun diff(from: MetamodelVersion, to: MetamodelVersion): MetamodelDiff {
+        val changes = mutableListOf<MetamodelChange>()
+
+        val fromTypes = from.entityTypeNames.toSet()
+        val toTypes = to.entityTypeNames.toSet()
+
+        val typePairing = pairRenamedTypes(
+            removed = fromTypes - toTypes,
```
igordayen, line 67: trying to understand the semantics of two lines: 67 and 68, as they are inverted

igordayen (commented):
@jimador - few minor comments about visibility.  Codex report - coming. thank you

igordayen:
from Codex:

1. The rename handling is broader for drift than it is for ```unobservedEntityTypes```

  Example:

  - Declared type name: Human
  - Declared aliases in the stamp: Person
  - Observed graph labels: Person

  Expected result:

  - This type should not be reported as unobserved, because the graph contains data under the declared former name of the same type.

  What the code does:

  - For drift, it treats declared former names as known, so Person is not reported as drift.
  - For ```unobservedEntityTypes```, it checks only whether the graph contains:
      - the current declared type name, or
      - ```ownLabelOf(currentDeclaredTypeName)```

  For this example, those checks are:

  - Human
  - Human

  They do not include the declared former name Person.

  Conclusion:

  - The same graph can be interpreted as “not drift” and also “type unobserved”.
  - _That means declared renames are only applied on one side of the declared-vs-observed comparison, not both_.

 

 2. The observed-side match is narrower than the declared schema it is comparing against.

  Example:

  - Declared type name: com.example.Outer$Person
  - Declared labels in the stamp: Person
  - Observed graph labels: Person

  Expected result:

  - This type should count as observed, because the graph contains the label the declaration says this type uses.

  What the code does:

  - For drift, it does consider the wider declared name set.
  - For unobservedEntityTypes, it checks only whether the graph contains:
      - the declared type name itself, or
      - ownLabelOf(declaredTypeName)

  For this example, those checks are:

  - com.example.Outer$Person
  - Outer$Person

  Neither matches the actual observed label Person.

  Conclusion:

  - A type can be reported as unobserved even when the graph contains data under the label declared for that type.
  - That means the observed-side matching rule is not using all of the declared identity information already present in the stamp.

jimador:
Answering the Codex findings.

Finding 1 is right and is fixed in `6897662` on this branch, with the stack above restacked on it.

1. `unobservedEntityTypes` now treats a declared former name as evidence the type is observed, the same way drift already does. `isObserved` takes the type's declared aliases and answers true when the current name or any former name is in the observed set, as spelled or as its own label. Three tests cover it: data only under the former name, data only under a qualified former name's label, and a guard that a type with no data under any of its names stays unobserved. The check widens to the type's own names only. A declared label set carries inherited parent labels, so accepting that as evidence would mark an empty `Person` observed whenever any `Agent` node exists.
2. The `Outer$Person` spelling is outside the built-in cut on purpose. The own-label KDoc names the two spellings the dot cut does not handle, a nested class with `$` and a platform that uppercases the first character, and says a host that meets either one maps its own names through the `TypeIdentity` SPI in this package. That SPI is a spec on this branch and its consuming slice lands separately, so this stays as is.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/StructuralMetamodelDiffer.kt
jimador: Yes. `dice-storage-autoconfigure` is a separate module and constructs it in `MetamodelAutoConfiguration`, and the design docs show a host constructing it directly when it wires the runner itself. Kotlin `internal` stops at the module boundary, so public is the narrowest visibility that works.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/StructuralMetamodelDiffer.kt
jimador: They read the way a diff does. `removed` is what `from` has and `to` lacks (`fromTypes - toTypes`), and `added` is what `to` has and `from` lacks (`toTypes - fromTypes`). The rename pairing then matches a removed name against an added type whose declared former name is that removed name.

CHANGELOG.md
jimador: Keeping it in the repo for now. It is a good reference for what is changing between releases, and it only moves between releases, so it will most likely get cleared out at release time.

igordayen (approved):
@jimador - looks good, thank you!

PR #86, opened 2026-08-30
feat(metamodel): drift checking and quarantine

jimador:
PR 4 of the metamodel train, stacked on #85. Drift checking and quarantine. `DriftCheckRunner` is a reporting-only contract: `run()` declares, stamps, observes, compares against the graph and against the swept baseline, and persists a `DriftReport` — its implementation holds no policy, no proposition store and no listener, so a check structurally cannot move a proposition. Acting on a check is a separate SPI, `DriftSweepCapable`: bounded, `ContextId`-scoped candidate paging, `applyQuarantine`, `releaseFromQuarantine`, and a defaulted `sweep` that persists what `DriftQuarantinePolicy` flags. Lossiness rules live in `MentionTypeDriftQuarantinePolicy`: type change, value/reference flip, or cardinality narrowing quarantines; widening never does; declared renames quarantine nothing by themselves, and former names are reachable however far a rename and a loss are separated. `DriftReport`/`DriftReportStore` carry the persistence contract with three explicitly scoped, bounded reads.

- **Quarantine has its own status.** `PropositionStatus.QUARANTINED` replaces the STALE-plus-metadata encoding, which the decay lifecycle used to undo: `DecayStatusPolicy` revived a quarantined proposition on the next sweep and the drift sweep re-held it, alternating forever. Now `DecayStatusPolicy` refuses `QUARANTINED` outright, reads filtering on `ACTIVE` exclude it structurally, `isAlreadyQuarantined` keys on status alone, and release is the only way out — restoring the recorded prior status, clearing the reason, emitting `PropositionStatusChanged`. A regression test proves three decay-then-drift rounds leave the hold in place.
- **The quarantine machinery moved to `dice`** (`com.embabel.dice.spi`, beside `StatusTransitionPolicy`), so `dice-metamodel` drops its `dice` dependency and is again a leaf over the agent `DataDictionary`.
- The runner backed off to report-only and the sweep became the explicit SPI above; declared-vs-previous (off the swept baseline, tracked apart from version history) now feeds quarantine, declared-vs-observed stays report-only.
- Pinned propositions keep their cross-cutting immunity: a lossy change reports them as `QuarantineDecision.Protected` and never flips them.
- The drift and lifecycle docs state the product rule once: a declared rename or alias changes how the diff and the drift check read old data; nothing rewrites a stored mention type.

**Breaking changes:** additive on the released surface, with three stated source-level exceptions — `QuarantineDecision` is sealed and gains `Protected`; `PropositionStatus` gains `QUARANTINED`, so an exhaustive `when` needs a branch (one internal site changed; a host that matches on status exhaustively recompiles); the quarantine types moved package before ever shipping. Persistence round-trips by enum name, so no stored value changes meaning.

**Opt-in and status:** **EXPERIMENTAL.** Nothing quarantines until a host calls `DriftSweepCapable.sweep()` on a diff it decided to act on; there is no scheduler and no default that moves a proposition.

```mermaid
flowchart TD
    A[DriftCheckRunner.run] --> B[diff declared vs previous swept baseline]
    A --> C[diff declared vs observed graph]
    B --> D[DriftReport, persisted]
    C --> D
    E[host calls DriftSweepCapable.sweep] --> F[DriftQuarantinePolicy]
    B --> F
    F --> G{pinned?}
    G -- yes --> H[Protected, reported and untouched]
    G -- no --> I{lossy change?}
    I -- yes --> J[QUARANTINED plus PropositionStatusChanged]
    I -- no --> K[untouched]
    J --> L[releaseFromQuarantine: prior status restored, reason cleared]
    M[decay sweep] -. cannot lift the hold .-> J
```

**Next in stack:** #87 Drivine drift persistence and observation, #88 wiring.

### DIFF
--- a/dice-metamodel/pom.xml
+++ b/dice-metamodel/pom.xml
@@ -15,9 +15,9 @@
     <dependencies>
         <!--
             Dice core: Proposition, PropositionStatus, PropositionRepository, DiceMetadataKeys.
-            Needed since quarantine landed — deciding a proposition is stranded and marking it
-            STALE means touching the proposition model, so the module can no longer be a leaf.
-            Anyone depending on dice-metamodel alone now pulls dice in transitively.
+            Needed since quarantine landed: marking a stranded proposition STALE touches the
+            proposition model, so this module is no longer a leaf. Anyone depending on
+            dice-metamodel alone now pulls dice in transitively.
         -->
         <dependency>
             <groupId>com.embabel.dice</groupId>
--- a/dice-metamodel/pom.xml
+++ b/dice-metamodel/pom.xml
@@ -14,17 +14,6 @@
 
     <dependencies>
         <!--
-            Dice core: Proposition, PropositionStatus, PropositionRepository, DiceMetadataKeys.
-            Needed since quarantine landed: marking a stranded proposition STALE touches the
-            proposition model, so this module is no longer a leaf. Anyone depending on
-            dice-metamodel alone now pulls dice in transitively.
-        -->
-        <dependency>
-            <groupId>com.embabel.dice</groupId>
-            <artifactId>dice</artifactId>
-        </dependency>
-
-        <!--
             Embabel agent core types (DataDictionary, DomainType, ContextId); provided, supplied by
             the consuming app.
         -->
--- a/dice-metamodel/pom.xml
+++ b/dice-metamodel/pom.xml
@@ -46,17 +46,6 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
-        <!--
-            https://mvnrepository.com/artifact/org.jetbrains/annotations
-            For Experimental annotation
-        -->
-        <dependency>
-            <groupId>org.jetbrains</groupId>
-            <artifactId>annotations</artifactId>
-            <version>26.0.2</version>
-            <scope>provided</scope>
-        </dependency>
-
         <!-- Test -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/dice-metamodel/pom.xml
+++ b/dice-metamodel/pom.xml
@@ -46,6 +46,17 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
+        <!--
+            https://mvnrepository.com/artifact/org.jetbrains/annotations
+            For Experimental annotation
+        -->
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>26.0.2</version>
+            <scope>provided</scope>
+        </dependency>
+
         <!-- Test -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/dice-metamodel/pom.xml
+++ b/dice-metamodel/pom.xml
@@ -10,13 +10,23 @@
     <artifactId>dice-metamodel</artifactId>
     <packaging>jar</packaging>
     <name>Dice Metamodel</name>
-    <description>Schema versioning for DICE knowledge graphs: content-hash stamping, the declared-schema contract, and the version store contract</description>
+    <description>Schema governance for DICE knowledge graphs: content-hash stamping, the declared-schema contract, diffing, drift checking, and non-destructive quarantine</description>
 
     <dependencies>
         <!--
-            Embabel agent core types (DataDictionary, DomainType); provided, supplied by the
-            consuming app. This module deliberately does not depend on dice core — it stamps a
-            schema and nothing more, so it never touches the Proposition model.
+            Dice core: Proposition, PropositionStatus, PropositionRepository, DiceMetadataKeys.
+            Needed since quarantine landed — deciding a proposition is stranded and marking it
+            STALE means touching the proposition model, so the module can no longer be a leaf.
+            Anyone depending on dice-metamodel alone now pulls dice in transitively.
+        -->
+        <dependency>
+            <groupId>com.embabel.dice</groupId>
+            <artifactId>dice</artifactId>
+        </dependency>
+
+        <!--
+            Embabel agent core types (DataDictionary, DomainType, ContextId); provided, supplied by
+            the consuming app.
         -->
         <dependency>
             <groupId>com.embabel.agent</groupId>
@@ -32,6 +42,19 @@
             <groupId>org.jetbrains</groupId>
             <artifactId>annotations</artifactId>
             <scope>provided</scope>
+        </dependency>
+
+        <!-- Retrievable, a supertype of Proposition, lives in rag-core and isn't transitive from dice -->
+        <dependency>
+            <groupId>com.embabel.agent</groupId>
+            <artifactId>embabel-agent-rag-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Logging only -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
         </dependency>
 
         <!-- Test -->
--- a/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
+++ b/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
@@ -21,16 +21,15 @@ import java.util.Objects
 /**
  * What one [DriftCheckRunner.run] call found and did.
  *
- * The drift itself lives in [report] and is read back off it, rather than being copied into
- * fields here. Every run persists a report, so the two would otherwise be two copies of the same
- * answer — and the failure mode of two copies is that they disagree, leaving a caller who logged
- * the result and an operator who read the stored report looking at different type sets for the same
- * check. There is one set of drifted types per run, and it is the one that got written down.
+ * The drifted types are read off [report] rather than copied into fields here. Every run persists a
+ * report, so a copy would be a second version of the same answer, and two versions can disagree: a
+ * caller who logged the result and an operator who read the stored report would then see different
+ * type sets for the same check.
  *
  * @property dryRun Whether this was a preview. On a dry run the report is still persisted; no
  *   proposition is touched.
- * @property report The [DriftReport] this run saved. Every run saves one, including a clean one —
- *   "checked and found nothing" has to be as retrievable as "checked and found drift".
+ * @property report The [DriftReport] this run saved. Every run saves one, including a check that
+ *   found nothing.
  * @property quarantinedCount How many propositions this run newly quarantined. Always 0 on a dry
  *   run, and 0 whenever there was no entity-type drift.
  */
@@ -69,25 +68,24 @@ class DriftCheckResult(
 
 /**
  * Runs a drift check end to end: takes the declared schema, stamps it, snapshots what a live graph
- * actually holds, compares the two, writes the result down, and — only if you ask — quarantines the
- * propositions the drift stranded.
+ * holds, compares the two, writes the result down, and quarantines the propositions the drift
+ * stranded when asked to.
  *
- * Dry-run by default, and that default is the whole design stance. Observing and reporting is
- * useful on its own and can't hurt anything; changing proposition state is a separate decision
- * somebody has to make on purpose. Nothing here schedules itself either — a consuming application
- * decides when [run] is called, the same way it does for the collector.
+ * Dry-run by default. Observing and reporting changes nothing; moving propositions to `STALE` is a
+ * separate decision a caller opts into. Nothing here schedules itself, so a consuming application
+ * decides when [run] is called, as it does for the collector.
  *
  * The shorter [run] forms are real overloads with bodies rather than Kotlin default arguments,
  * because Java can't see a default argument: `runner.run()` has to exist as a method for a Java
  * caller to write it. Implementations override the two-argument form and get the other two free.
- * Those two shorter forms are also the whole Java surface — `ContextId` is a Kotlin value class, so
- * the two-argument form compiles to a mangled JVM name Java can't call.
+ * Those two shorter forms are also the whole Java surface, since `ContextId` is a Kotlin value class
+ * and the two-argument form compiles to a mangled JVM name Java can't call.
  */
 interface DriftCheckRunner {
 
     /**
-     * Declare, stamp, observe, diff, report — and, when [dryRun] is `false` and drift touched any
-     * entity type, quarantine.
+     * Declare, stamp, observe, diff, report, and quarantine when [dryRun] is `false` and drift
+     * touched an entity type.
      *
      * @param dryRun When `true`, the check runs and its [DriftReport] is persisted, but no
      *   proposition is touched. When `false`, propositions whose mentions reference a drifted
@@ -96,15 +94,13 @@ interface DriftCheckRunner {
      * @param contextId `null` means the check covers the whole graph. Non-null scopes everything
      *   the check touches to that one context: the observed snapshot, the candidate propositions
      *   read for quarantine, and the persisted [DriftReport]. A mis-declared schema in one context
-     *   can then only ever quarantine propositions in that same context — it has no way to reach
-     *   another one.
+     *   can only quarantine propositions in that same context.
      * @return What was found, and what was quarantined if this was a live run.
      */
     fun run(dryRun: Boolean, contextId: ContextId?): DriftCheckResult
 
     /**
-     * Run a dry check over the whole graph — the safe default. Nothing is quarantined; the report
-     * is still persisted.
+     * Run a dry check over the whole graph. Nothing is quarantined; the report is still persisted.
      *
      * @return What was found.
      */
--- a/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
+++ b/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2024-2026 Embabel Pty Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.metamodel
+
+import com.embabel.agent.core.ContextId
+import java.util.Objects
+
+/**
+ * What one [DriftCheckRunner.run] call found and did.
+ *
+ * The drift itself lives in [report] and is read back off it, rather than being copied into
+ * fields here. Every run persists a report, so the two would otherwise be two copies of the same
+ * answer — and the failure mode of two copies is that they disagree, leaving a caller who logged
+ * the result and an operator who read the stored report looking at different type sets for the same
+ * check. There is one set of drifted types per run, and it is the one that got written down.
+ *
+ * @property dryRun Whether this was a preview. On a dry run the report is still persisted; no
+ *   proposition is touched.
+ * @property report The [DriftReport] this run saved. Every run saves one, including a clean one —
+ *   "checked and found nothing" has to be as retrievable as "checked and found drift".
+ * @property quarantinedCount How many propositions this run newly quarantined. Always 0 on a dry
+ *   run, and 0 whenever there was no entity-type drift.
+ */
+class DriftCheckResult(
+    val dryRun: Boolean,
+    val report: DriftReport,
+    val quarantinedCount: Int,
+) {
+
+    /** The declared schema the check ran against. */
+    val schemaName: String get() = report.schemaName
+
+    /** The context the check was scoped to, or `null` when it covered the whole graph. */
+    val contextId: ContextId? get() = report.contextId
+
+    /** Entity type names the graph held but the schema never declared. */
+    val driftedEntityTypes: Set<String> get() = report.driftedEntityTypes
+
+    /** Relationship type names observed with no matching declaration. */
+    val driftedRelationshipTypes: Set<String> get() = report.driftedRelationshipTypes
+
+    /** `true` when the graph contained any type or relationship that was never declared. */
+    val hasDrift: Boolean get() = report.hasDrift
+
+    override fun equals(other: Any?): Boolean =
+        other is DriftCheckResult &&
+            dryRun == other.dryRun &&
+            report == other.report &&
+            quarantinedCount == other.quarantinedCount
+
+    override fun hashCode(): Int = Objects.hash(dryRun, report, quarantinedCount)
+
+    override fun toString(): String =
+        "DriftCheckResult(dryRun=$dryRun, quarantinedCount=$quarantinedCount, report=$report)"
+}
+
+/**
+ * Runs a drift check end to end: takes the declared schema, stamps it, snapshots what a live graph
+ * actually holds, compares the two, writes the result down, and — only if you ask — quarantines the
+ * propositions the drift stranded.
+ *
+ * Dry-run by default, and that default is the whole design stance. Observing and reporting is
+ * useful on its own and can't hurt anything; changing proposition state is a separate decision
+ * somebody has to make on purpose. Nothing here schedules itself either — a consuming application
+ * decides when [run] is called, the same way it does for the collector.
+ *
+ * The shorter [run] forms are real overloads with bodies rather than Kotlin default arguments,
+ * because Java can't see a default argument: `runner.run()` has to exist as a method for a Java
+ * caller to write it. Implementations override the two-argument form and get the other two free.
+ * Those two shorter forms are also the whole Java surface — `ContextId` is a Kotlin value class, so
+ * the two-argument form compiles to a mangled JVM name Java can't call.
+ */
+interface DriftCheckRunner {
+
+    /**
+     * Declare, stamp, observe, diff, report — and, when [dryRun] is `false` and drift touched any
+     * entity type, quarantine.
+     *
+     * @param dryRun When `true`, the check runs and its [DriftReport] is persisted, but no
+     *   proposition is touched. When `false`, propositions whose mentions reference a drifted
+     *   entity type are handed to the configured [DriftQuarantinePolicy] and whatever it flags is
+     *   persisted.
+     * @param contextId `null` means the check covers the whole graph. Non-null scopes everything
+     *   the check touches to that one context: the observed snapshot, the candidate propositions
+     *   read for quarantine, and the persisted [DriftReport]. A mis-declared schema in one context
+     *   can then only ever quarantine propositions in that same context — it has no way to reach
+     *   another one.
+     * @return What was found, and what was quarantined if this was a live run.
+     */
+    fun run(dryRun: Boolean, contextId: ContextId?): DriftCheckResult
+
+    /**
+     * Run a dry check over the whole graph — the safe default. Nothing is quarantined; the report
+     * is still persisted.
+     *
+     * @return What was found.
+     */
+    fun run(): DriftCheckResult = run(dryRun = true, contextId = null)
+
+    /**
+     * Run over the whole graph, dry or live.
+     *
+     * @param dryRun `true` to preview without touching any proposition.
+     * @return What was found, and what was quarantined if this was a live run.
+     */
+    fun run(dryRun: Boolean): DriftCheckResult = run(dryRun, null)
+}
--- a/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
+++ b/dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
@@ -19,27 +19,30 @@ import com.embabel.agent.core.ContextId
 import java.util.Objects
 
 /**
- * What one [DriftCheckRunner.run] call found and did.
+ * What one [DriftCheckRunner.run] call found.
  *
- * The drifted types are read off [report] rather than copied into fields here. Every run persists a
+ * The drifted types are read off [report], with no second copy kept here. Every run persists a
  * report, so a copy would be a second version of the same answer, and two versions can disagree: a
  * caller who logged the result and an operator who read the stored report would then see different
  * type sets for the same check.
  *
- * @property dryRun Whether this was a preview. On a dry run the report is still persisted; no
- *   proposition is touched.
  * @property report The [DriftReport] this run saved. Every run saves one, including a check that
  *   found nothing.
- * @property quarantinedCount How many propositions this run newly quarantined. Always 0 on a dry
- *   run. On a live run it can be non-zero even when [driftedEntityTypes] and
- *   [driftedRelationshipTypes] are both empty: quarantine reacts to two independent signals, not
- *   just observed drift — see [DriftCheckRunner.run]'s `dryRun` parameter doc for the second one.
+ * @property declaredVersion The stamp the check ran against. [DriftReport.versionHash] is this
+ *   stamp's [MetamodelVersion.contentHash], and holding the stamp itself is what lets [quarantineDiff]
+ *   answer without a second trip to the version store.
  */
 class DriftCheckResult(
-    val dryRun: Boolean,
     val report: DriftReport,
-    val quarantinedCount: Int,
+    val declaredVersion: MetamodelVersion,
 ) {
+
+    init {
+        require(report.versionHash == declaredVersion.contentHash) {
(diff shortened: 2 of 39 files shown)


dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
```diff
@@ -0,0 +1,157 @@
...
+) : DriftCheckRunner {
+
+    private val logger = LoggerFactory.getLogger(DefaultDriftCheckRunner::class.java)
+
+    override fun run(dryRun: Boolean, contextId: ContextId?): DriftCheckResult {
+        val declared = declaredSchemaSource.declare()
+
+        // Before the report is written, so the hash the report carries is already resolvable by the
+        // time anyone can read it. See the class doc.
+        versionStore.saveVersion(declared.version)
```
jimador, line 124: Where is the previous stored version compared before this one becomes current? This runner only does declared-versus-observed name drift, so property removals and type or cardinality changes never reach quarantine.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/MentionTypeDriftQuarantinePolicy.kt
```diff
@@ -0,0 +1,454 @@
...
+                removedTypes = removedHit,
+                lossyChanges = lossyHit.toList(),
+                narrowedChanges = narrowedHit.toList(),
+                renamedChanges = renamedHit.toList(),
+                matchedByFormerName = matchedByFormerName,
+                fromSchema = diff.fromVersion.schemaName,
+                toSchema = diff.toVersion.schemaName,
+            )
+            val flagged = proposition
+                .withStatus(PropositionStatus.STALE)
```
jimador: Pinned propositions reach this status change and lose the cross-cutting immunity DICE promises for protected knowledge. Can we skip them here, with the result reporting that they were protected?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
```diff
@@ -0,0 +1,157 @@
...
+        )
+        // A proposition in another context is never a candidate, whatever its mentions say, so a
+        // scoped run cannot reach outside its context.
+        val propositions = if (contextId != null) {
+            propositionStore.findByContextId(contextId)
+        } else {
+            propositionStore.findAll()
+        }
+        val result = quarantinePolicy.evaluate(syntheticDiff, propositions)
+        result.quarantined.forEach { decision -> propositionStore.save(decision.proposition) }
```
jimador: This save bypasses `PropositionStatusChanged`, so with the default raw repository wiring projection lineage stays active after the proposition becomes stale. Can we route quarantine through the event-emitting path?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
jimador: Added the second comparison. `sweptVersion` supplies the declared-vs-previous baseline, read before the check, and `metamodelDiffer.diff(previous, declared)` feeds quarantine, so property removals and type or cardinality changes reach it. Declared-vs-observed stays report-only.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/MentionTypeDriftQuarantinePolicy.kt
jimador: Pinned propositions are skipped and reported. A pinned match a lossy change would have caught comes back as `QuarantineDecision.Protected` carrying the reason, so an operator can see what the schema change would have quarantined.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
jimador: The runner emits `PropositionStatusChanged` for each quarantine through an injected listener, so projection lineage sees the status change under the default raw repository wiring.

jimador:
Three issues in the runner/store contract need resolving together:

- A dry run can report clean immediately before the same state quarantines on a live run. `declaredDiff` is computed but omitted from `DriftReport` and `DriftCheckResult`. With no completed sweep, dry A → dry B → live B can also lose the only A→B comparison entirely.
- `MetamodelVersionStore.sweptVersion` defaults to `latestVersion`, and `markSwept` defaults to `saveVersion`. A custom store implementing the original contract still compiles but silently treats dry, scoped, or interrupted history writes as completed sweeps. This state needs a required capability or abstract methods.
- An unscoped live run calls `PropositionStore.findAll()`, materializes every tenant, and filters mention types in the JVM. That violates the ContextId-first/store-pushdown rules and makes one drifted type proportional to the whole corpus.

A dry run should evaluate and return the same candidate impact as live without changing propositions or the completed-sweep baseline. Live candidate selection needs a bounded, context/type-aware store operation; it cannot be expressed through `findAll`.

The reversibility claim also needs a real release operation. Clearing `dice.metamodel.quarantineReason` leaves the proposition `STALE`, so it remains excluded from normal retrieval and projection.

jimador:
Quarantine as built is a status value with no owner, and the existing lifecycle undoes it.

`MentionTypeDriftQuarantinePolicy.kt:253-254` sets `STALE` plus `QUARANTINE_REASON`. The default `DecayStatusPolicy` (`StatusTransitionPolicy.kt:99-111` on main) moves any `STALE` proposition back to `ACTIVE` when utility is above `recoveryThreshold`, with no check of the reason key, and `DecayManager.kt:82-84` saves that transition on every sweep. A freshly quarantined proposition still has the confidence that kept it active, so the next decay sweep revives it. `withStatus` (`Proposition.kt:254`) keeps metadata, so it comes back `ACTIVE` with the quarantine reason still attached. `isAlreadyQuarantined` (`MentionTypeDriftQuarantinePolicy.kt:294-296`) requires `STALE`, so the next live drift run quarantines it again. The two sweeps alternate. The positioning doc describes quarantine as stored, excluded from reads and recoverable. As shipped, recovery is accidental.

The fix is a disposition the lifecycle honours. Two shapes:

- A `PropositionStatus.QUARANTINED` value. Reads that filter on `ACTIVE` exclude it for free, `DecayStatusPolicy` never sees a `STALE` to revive, and release is an explicit transition. Cost is an enum addition. `me` matches on status in `MemorySearchQueryParser.kt:68`, and the stack already accepted an exhaustive-`when` break in #85, so the boundary is known.
- Keep `STALE` and make the lifecycle policies treat `QUARANTINE_REASON` the way they treat `pinned`. `StatusTransitionPolicy`, `DecayManager`, the collector sweep and contradiction resolution each check `pinned` by hand, and this policy re-implements the pin check at `:239`. That is the second cross-cutting immunity implemented per policy. If we go this way, one guard in `dice` that both quarantine and pin route through.

I would take the status. Either way the release operation has to exist as a call (clear reason, restore `ACTIVE`, emit `PropositionStatusChanged`), and the thread above already asks for it.

Related. `dice-metamodel` now depends on `dice` because quarantine touches `Proposition` (`dice-metamodel/pom.xml:22-25`). Versioning and diff are a leaf library over `DataDictionary` and work without the proposition model. Quarantine is a proposition lifecycle policy and belongs next to the other lifecycle policies in `dice`. Splitting it out keeps the leaf and puts the disposition where `StatusTransitionPolicy` lives, which is where the conflict above gets resolved anyway.

One product expectation to state once in the design doc and the positioning doc: declared renames and aliases change how the diff and the drift check read old data. Nothing rewrites a stored mention type.

jimador:
Took the status. Landed as "Give quarantine its own status and move it beside the lifecycle policies", with the doc pass in "Align the drift and lifecycle docs with the shipped model".

**The alternation is closed structurally.** `PropositionStatus.QUARANTINED` is its own value. `DecayStatusPolicy` returns null for it outright — an explicit guard, so "lifecycle policies ignore quarantined propositions" is a stated rule with a test on it. Reads that filter on `ACTIVE` exclude it for free; `isAlreadyQuarantined` keys on status alone, so a hand-edited reason key can no longer release a hold. A regression test runs three decay-then-drift rounds with `targetStatuses` widened to every status and asserts the hold survives each; reverting the guard fails three of its assertions, including the alternation itself. The exhaustive-`when` boundary you named is the one accepted in #85: one internal site (`DefaultDreamLoopOrchestrator.statusStrength`, where `QUARANTINED` now ranks above `CONTRADICTED`) and `me` recompiles. Persistence round-trips by enum name, so no stored value changes meaning.

**Release exists as a call.** `releaseFromQuarantine` keys on status, restores the recorded prior status, clears the reason, and emits `PropositionStatusChanged`. Releasing twice finds a non-quarantined proposition and answers null. The operator-facing call sits on #88.

**The leaf is restored.** The same commit moves `DriftQuarantinePolicy`, `QuarantineDecision`, `QuarantineResult`, `MentionTypeDriftQuarantinePolicy`, `DriftSweepCapable` and `PropositionStoreDriftSweep` into `com.embabel.dice.spi`, beside `StatusTransitionPolicy` — which is where the decay conflict got resolved. `dice-metamodel` drops its `dice` dependency and is again a leaf over the `DataDictionary`; `dice` gains a dependency on `dice-metamodel` to read the diff. Pure package move on an unreleased module.

**The product line is stated once**, in `docs/design/metamodel-drift.md`: a declared rename or alias changes how the diff and the drift check read old data; nothing rewrites a stored mention type. The positioning doc makes no claim about schema evolution or stored mention types, so it needed no edit — adding the line there would have been new claim-making.

jimador:
Status and package move are right, and the `DecayStatusPolicy` guard closes the alternation. One doc point. `docs/design/proposition-lifecycle.md` draws only `ACTIVE --> QUARANTINED`, while `DriftQuarantinePolicy` says a proposition can be quarantined from any non-quarantined status, including `STALE`, and `releaseFromQuarantine` restores the recorded prior status rather than `ACTIVE`. Draw both, and add the QUARANTINED transitions to the trigger list under the diagram the way every other transition has one.

james-gsb:
Redrawn on ec7b957. The diagram now has QUARANTINED edges from ACTIVE, STALE, CONTRADICTED and SUPERSEDED, matching `DriftQuarantinePolicy.evaluate`, and the release edges restore the recorded prior status from `DriftQuarantineKeys.PREVIOUS_STATUS`. Every new transition has its trigger-list entry under the diagram in the established format.

Same commit marks the drift surface: `@ApiStatus.Experimental` on `DriftCheckRunner`, `DriftCheckResult`, `DriftReport`, `DriftReportStore`, `DriftQuarantinePolicy`, `DriftQuarantineKeys`, `QuarantineResult`, `DriftSweepCapable`, `PropositionStoreDriftSweep` and `MentionTypeDriftQuarantinePolicy`.

jimador:
Redrawn on ec7b957. The diagram now has QUARANTINED edges from ACTIVE, STALE, CONTRADICTED and SUPERSEDED, matching `DriftQuarantinePolicy.evaluate`, and the release edges restore the recorded prior status from `DriftQuarantineKeys.PREVIOUS_STATUS`. Every new transition has its trigger-list entry under the diagram in the established format.

Same commit marks the drift surface: `@ApiStatus.Experimental` on `DriftCheckRunner`, `DriftCheckResult`, `DriftReport`, `DriftReportStore`, `DriftQuarantinePolicy`, `DriftQuarantineKeys`, `QuarantineResult`, `DriftSweepCapable`, `PropositionStoreDriftSweep` and `MentionTypeDriftQuarantinePolicy`.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftReport.kt
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * [versionHash] is the tie back to the schema this was judged against. A `DriftCheckRunner` stamps
+ * the declared version into a [MetamodelVersionStore] before writing the report, so the hash always
+ * resolves through [MetamodelVersionStore.findVersion]. Pull a year-old report and you can still
+ * recover the exact shape that was expected when it was taken.
+ *
+ * ## A report carries both halves of the comparison
+ *
+ * [driftedEntityTypes] and [driftedRelationshipTypes] are the graph-truth half: what the graph holds
+ * that this declaration doesn't recognise. [declaredDiff] is the other half: how the declaration
```
jimador, line 40: A dry run can report clean immediately before the same state quarantines on a live run. `declaredDiff` is computed but omitted from `DriftReport` and `DriftCheckResult`. With no completed sweep, dry A → dry B → live B can also lose the only A→B comparison entirely. A dry run should evaluate and return the same candidate impact as live without changing propositions or the completed-sweep baseline.

jimador: Dry reports carry `declaredDiff`: the runner reads the swept baseline and feeds declared-vs-previous to the report, so a dry run answers the full impact a sweep would evaluate. The runner is report-only by construction; quarantine happens only when a host calls `DriftSweepCapable.sweep` on purpose.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/MetamodelVersionStore.kt
```diff
@@ -89,3 +89,66 @@ interface MetamodelVersionStore {
...
+     * The version the last COMPLETED drift sweep reconciled against — the baseline for the next
+     * declared-vs-previous comparison.
+     *
+     * This is a different question from [latestVersion], which answers "what's the newest stamp by
+     * write order" and gets the wrong answer once a declaration cycles back to a stamp that already
+     * exists: [saveVersion]'s own contract says a re-saved stamp keeps its *original* place in write
+     * order, so after a schema goes `A` → `B` → `A` again, [latestVersion] still answers `B`, even
+     * though the schema is back to declaring `A`. A drift check that diffed against [latestVersion]
+     * would compare the reverted `A` against `B` — the wrong pair — and could miss a lossy change
+     * that came back. [sweptVersion] tracks the reconciled baseline itself, moved forward only by
```
jimador, line 135: `MetamodelVersionStore.sweptVersion` defaults to `latestVersion`, and `markSwept` defaults to `saveVersion`. A custom store implementing the original contract still compiles but silently treats dry, scoped, or interrupted history writes as completed sweeps. This state needs a required capability or abstract methods.

jimador: The swept baseline is its own capability: a store that can hold a completed-sweep marker declares `SweptBaselineStore` and binds `markSwept`/`sweptVersion` itself. `DrivineMetamodelVersionStore` declares it on #87. A store that never mentions it cannot be mistaken for one that swept.

dice/src/main/kotlin/com/embabel/dice/spi/DriftSweepCapable.kt
```diff
@@ -0,0 +1,249 @@
...
+     * finished. See that method for what goes wrong if the baseline moves early.
+     *
+     * @param diff The comparison to evaluate against, ordinarily [DriftCheckResult.quarantineDiff].
+     * @param policy Decides which propositions the change stranded.
+     * @param contextId The context to sweep.
+     * @param batchSize How many candidates to read per page. Must be positive.
+     * @return Every decision the sweep made, gathered across pages in the order they were read.
+     * @throws IllegalArgumentException if [batchSize] is not positive.
+     */
+    fun sweep(
```
jimador, line 202: An unscoped live run calls `PropositionStore.findAll()`, materializes every tenant, and filters mention types in the JVM. That violates the ContextId-first/store-pushdown rules and makes one drifted type proportional to the whole corpus. Live candidate selection needs a bounded, context/type-aware store operation; it cannot be expressed through `findAll`.

jimador: Live selection is a store operation now. `DriftSweepCapable.sweep` is the SPI, `PropositionStoreDriftSweep` the default over a store's own context- and type-aware read, and live drift mode is backed off to that deliberate sweep call ("Back live drift mode off to a deliberate sweep SPI"). `findAll` is no longer on the path.

dice/src/main/kotlin/com/embabel/dice/spi/DriftQuarantinePolicy.kt
```diff
@@ -0,0 +1,237 @@
...
+@ApiStatus.Experimental
+object DriftQuarantineKeys {
+
+    /**
+     * The [PropositionStatus] a proposition carried at the moment it was quarantined, stored as its
+     * `name`.
+     *
+     * A proposition can be quarantined from any status — active, promoted, already stale from
+     * ordinary decay — so without this key, releasing one could only guess where to put it back.
+     * With it, release is exact: [DriftSweepCapable.releaseFromQuarantine] reads the value, restores
```
jimador, line 41: The reversibility claim also needs a real release operation. Clearing `dice.metamodel.quarantineReason` leaves the proposition `STALE`, so it remains excluded from normal retrieval and projection.

jimador: `releaseFromQuarantine` exists as a call: it keys on status, restores the recorded prior status, clears the reason, and emits `PropositionStatusChanged`. Releasing twice finds a non-quarantined proposition and answers null. The operator-facing call sits on #88.

dice/src/main/kotlin/com/embabel/dice/proposition/Proposition.kt
```diff
@@ -50,7 +50,23 @@ enum class PropositionStatus {
...
+     * owns it: decay reaches [STALE] on its own and can lift a proposition back out on its own,
+     * while nothing automatic touches a quarantined proposition. Lifecycle policies leave it alone
+     * — no decay transition, no revival, no contradiction-resolution move — and the one way out is
+     * an explicit release, which restores the status the proposition carried before quarantine.
+     *
+     * The reason it was held is on the proposition, under
+     * `DiceMetadataKeys.QUARANTINE_REASON`.
+     */
+    @ApiStatus.Experimental
+    QUARANTINED
```
jimador, line 69: Quarantine as built is a status value with no owner, and the existing lifecycle undoes it.

`MentionTypeDriftQuarantinePolicy` sets `STALE` plus `QUARANTINE_REASON`. The default `DecayStatusPolicy` (`StatusTransitionPolicy.kt` on main) moves any `STALE` proposition back to `ACTIVE` when utility is above `recoveryThreshold`, with no check of the reason key, and `DecayManager` saves that transition on every sweep. A freshly quarantined proposition still has the confidence that kept it active, so the next decay sweep revives it. `withStatus` keeps metadata, so it comes back `ACTIVE` with the quarantine reason still attached. `isAlreadyQuarantined` requires `STALE`, so the next live drift run quarantines it again. The two sweeps alternate. The positioning doc describes quarantine as stored, excluded from reads and recoverable. As shipped, recovery is accidental.

The fix is a disposition the lifecycle honours. Two shapes:

- A `PropositionStatus.QUARANTINED` value. Reads that filter on `ACTIVE` exclude it for free, `DecayStatusPolicy` never sees a `STALE` to revive, and release is an explicit transition. Cost is an enum addition. A host that matches on status exhaustively takes a `when` break, and the stack already accepted one in #85, so the boundary is known.
- Keep `STALE` and make the lifecycle policies treat `QUARANTINE_REASON` the way they treat `pinned`. `StatusTransitionPolicy`, `DecayManager`, the collector sweep and contradiction resolution each check `pinned` by hand, and this policy re-implements the pin check. That is the second cross-cutting immunity implemented per policy. If we go this way, one guard in `dice` that both quarantine and pin route through.

I would take the status. Either way the release operation has to exist as a call (clear reason, restore `ACTIVE`, emit `PropositionStatusChanged`).

jimador: Took the status. Landed as "Give quarantine its own status and move it beside the lifecycle policies", with the doc pass in "Align the drift and lifecycle docs with the shipped model".

**The alternation is closed structurally.** `PropositionStatus.QUARANTINED` is its own value. `DecayStatusPolicy` returns null for it outright, an explicit guard, so "lifecycle policies ignore quarantined propositions" is a stated rule with a test on it. Reads that filter on `ACTIVE` exclude it for free; `isAlreadyQuarantined` keys on status alone, so a hand-edited reason key can no longer release a hold. A regression test runs three decay-then-drift rounds with `targetStatuses` widened to every status and asserts the hold survives each; reverting the guard fails three of its assertions, including the alternation itself. The exhaustive-`when` boundary is the one accepted in #85: one internal site (`DefaultDreamLoopOrchestrator.statusStrength`, where `QUARANTINED` now ranks above `CONTRADICTED`) and a host that matches on status recompiles. Persistence round-trips by enum name, so no stored value changes meaning.

**Release exists as a call.** `releaseFromQuarantine` keys on status, restores the recorded prior status, clears the reason, and emits `PropositionStatusChanged`. Releasing twice finds a non-quarantined proposition and answers null. The operator-facing call sits on #88.

dice/pom.xml
```diff
@@ -26,6 +26,18 @@
             <scope>provided</scope>
         </dependency>
 
+        <!--
+            Schema governance: MetamodelDiff, MetamodelChange, DeclaredSchema, PropertySignature.
+            The drift quarantine policy and the sweep contract live here in dice core, beside the
+            other proposition lifecycle policies, and they read a diff to decide what a schema
+            change stranded. dice-metamodel itself is a leaf over the agent DataDictionary and
```
jimador, line 33: `dice-metamodel` now depends on `dice` because quarantine touches `Proposition`. Versioning and diff are a leaf library over `DataDictionary` and work without the proposition model. Quarantine is a proposition lifecycle policy and belongs next to the other lifecycle policies in `dice`. Splitting it out keeps the leaf and puts the disposition where `StatusTransitionPolicy` lives, which is where the decay conflict gets resolved anyway.

jimador: **The leaf is restored.** The same commit moves `DriftQuarantinePolicy`, `QuarantineDecision`, `QuarantineResult`, `MentionTypeDriftQuarantinePolicy`, `DriftSweepCapable` and `PropositionStoreDriftSweep` into `com.embabel.dice.spi`, beside `StatusTransitionPolicy`, which is where the decay conflict got resolved. `dice-metamodel` drops its `dice` dependency and is again a leaf over the `DataDictionary`; `dice` gains a dependency on `dice-metamodel` to read the diff. Pure package move on an unreleased module.

docs/design/metamodel-drift.md
```diff
@@ -0,0 +1,643 @@
...
+the old ones, and guessing wrong in the permissive direction leaves unreadable data looking healthy.
+
+### Declared renames
+
+A rename is a fact the declaration states, so on its own it strands nothing.
+`EntityTypeRenamed` and `PropertyRenamed` are non-lossy per se, and `EntityTypeAliasesChanged` never
+quarantines at all.
+
+**A declared rename or alias changes how the diff and the drift check read old data. Nothing rewrites
+a stored mention type.** A proposition extracted under `Person` still says `Person` after the schema
```
jimador, line 353: One product expectation to state once in the design doc and the positioning doc: declared renames and aliases change how the diff and the drift check read old data. Nothing rewrites a stored mention type.

jimador: **The product line is stated once**, here: a declared rename or alias changes how the diff and the drift check read old data; nothing rewrites a stored mention type. The positioning doc makes no claim about schema evolution or stored mention types, so it needed no edit; adding the line there would have been new claim-making.

docs/design/proposition-lifecycle.md
```diff
@@ -19,21 +19,39 @@ stateDiagram-v2
     ACTIVE --> CONTRADICTED : a newer fact clashes with it (revision or ContradictionResolutionPass)
     ACTIVE --> SUPERSEDED : folded into a higher-level abstraction (AbstractionPass)
     ACTIVE --> STALE : DecayStatusPolicy.evaluate — utility drops below stalenessThreshold
+    ACTIVE --> QUARANTINED : DriftSweepCapable.sweep — schema drift detected (see metamodel-drift.md)
```
jimador, line 22: Status and package move are right, and the `DecayStatusPolicy` guard closes the alternation. One doc point. This diagram draws only `ACTIVE --> QUARANTINED`, while `DriftQuarantinePolicy` says a proposition can be quarantined from any non-quarantined status, including `STALE`, and `releaseFromQuarantine` restores the recorded prior status, not `ACTIVE`. Draw both, and add the QUARANTINED transitions to the trigger list under the diagram the way every other transition has one.

jimador: Redrawn on ec7b957. The diagram now has QUARANTINED edges from ACTIVE, STALE, CONTRADICTED and SUPERSEDED, matching `DriftQuarantinePolicy.evaluate`, and the release edges restore the recorded prior status from `DriftQuarantineKeys.PREVIOUS_STATUS`. Every new transition has its trigger-list entry under the diagram in the established format.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/DriftCheckRunner.kt
```diff
@@ -0,0 +1,137 @@
...
+ * caller who logged the result and an operator who read the stored report would then see different
+ * type sets for the same check.
+ *
+ * @property report The [DriftReport] this run saved. Every run saves one, including a check that
+ *   found nothing.
+ * @property declaredVersion The stamp the check ran against. [DriftReport.versionHash] is this
+ *   stamp's [MetamodelVersion.contentHash], and holding the stamp itself is what lets [quarantineDiff]
+ *   answer without a second trip to the version store.
+ */
+@ApiStatus.Experimental
```
jimador, line 37: The drift surface carried no `@ApiStatus.Experimental` while the CHANGELOG called it experimental in prose. Mark the public types.

jimador: Marked on ec7b957: `DriftCheckRunner`, `DriftCheckResult`, `DriftReport`, `DriftReportStore`, `DriftQuarantinePolicy`, `DriftQuarantineKeys`, `QuarantineResult`, `DriftSweepCapable`, `PropositionStoreDriftSweep` and `MentionTypeDriftQuarantinePolicy`.

igordayen:
From CODEX:

1. dice/src/main/kotlin/com/embabel/dice/spi/MentionTypeDriftQuarantinePolicy.kt:452

  This can miss propositions that should be quarantined.

  Example:

  - schema type: com.example.Outer$Person
  - stored mention type: Person
  - later, that type loses the property age

  That proposition should be quarantined because it refers to the type that changed.

  But this policy only recognizes the type as:

  - com.example.Outer$Person
  - Outer$Person

  It does not recognize Person as the same type. So when a proposition is stored under Person, this policy treats it as unrelated to
  the changed type and leaves it unquarantined.

  Conclusion:
  Lossy changes can be missed when the stored label differs from the names this policy checks.

dice-metamodel/pom.xml
```diff
@@ -34,6 +33,30 @@
...
 
+        <!-- Retrievable, a supertype of Proposition, lives in rag-core and isn't transitive from dice -->
+        <dependency>
+            <groupId>com.embabel.agent</groupId>
+            <artifactId>embabel-agent-rag-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Logging only -->
+        <dependency>
```
igordayen, line 44: rerquired? not transitive?

dice-metamodel/pom.xml
```diff
@@ -34,6 +33,30 @@
...
+        </dependency>
+
+        <!--
+            https://mvnrepository.com/artifact/org.jetbrains/annotations
+            For Experimental annotation
+        -->
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>26.0.2</version>
```
igordayen: should be inherited from embabel-agent. also please consider version meta-property, thanks

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
```diff
@@ -0,0 +1,160 @@
...
+ *   is judged against one declaration.
+ * @param versionStore Where the declared stamp is recorded each run, so report hashes always
+ *   resolve, and where the reconciled baseline is read from when the store tracks one.
+ * @param observedSchemaSource Snapshots what the live graph actually contains.
+ * @param differ Compares the declaration against the observation.
+ * @param metamodelDiffer Compares the declaration against its reconciled baseline. The same
+ *   [StructuralMetamodelDiffer] instance ordinarily implements both this and [differ].
+ * @param driftReportStore Durable log the report is written to, on every run.
+ */
+class DefaultDriftCheckRunner(
```
igordayen, line 103: public visibility?

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
```diff
@@ -0,0 +1,160 @@
...
+            // report is a statement about the snapshot.
+            capturedAt = observed.capturedAt,
+            contextId = contextId,
+            declaredDiff = declaredDiff,
+        )
+        // Written on every run, including checks that found nothing.
+        driftReportStore.saveDriftReport(report)
+
+        logger.info(
+            "Drift check for '{}' complete (contextId={}): {} drifted entity type(s), " +
```
igordayen, line 149: multiline string

igordayen:
From CODEX - inquiry on concurrency:

2. dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/InMemoryMetamodelVersionStore.kt:69

  markSwept() is last-write-wins with no guard against an older sweep finishing after a newer one, so the baseline can move backward.

  Example:

  - sweep 1 starts for schema version A
  - schema changes, then sweep 2 starts for newer version B
  - sweep 2 finishes first and calls markSwept(B)
  - sweep 1 finishes later and calls markSwept(A)

  After that, sweptVersion() returns A again even though B was already reconciled.

  Why that matters:
  The next drift check compares the current declaration against A, so it can re-report changes from A -> B that were already swept. In
  other words, an older concurrent sweep can overwrite the newer baseline just by finishing later.

  markSwept() needs some way to reject stale completions, or the caller has to guarantee sweeps for one schema never overlap.

igordayen (commented):
@jimador - few comments from me and Codex:)  thank you

jimador:
On Codex finding 1:

The `Outer$Person` case is the nested-class spelling that sits outside the built-in dot cut, and it is handled the same way in the quarantine policy as in the differ on #85: the own-label KDoc names it as one of the two spellings a host maps itself through the `TypeIdentity` SPI. The policy's `spellingsOf` uses the same cut as the differ, so the two stay in agreement, and the consuming slice for `TypeIdentity` lands separately.

On finding 2 (concurrency):

Correct: `markSwept` is last-write-wins and DICE never calls it. The host does, after a sweep. The contract is now written on the interface in `8c2683b`: sweeps of one schema must not overlap, and the call site runs them one at a time. A compare-and-set form, `markSwept(version, expectedBaseline)`, is the follow-up if a host ever needs overlapping sweeps. I would take it as a separate issue.

dice-metamodel/pom.xml
jimador: Required. `DefaultDriftCheckRunner` logs through slf4j directly, and an API a module calls itself is declared in that module. `dice-storage` and `dice-storage-autoconfigure` declare it the same way.

dice-metamodel/pom.xml
jimador: Fixed in `8c2683b` and `fbed94e`. The block with the hard-coded `26.0.2` was a leftover, and the same commit had added one to `dice/pom.xml`, so both are gone. The declaration left in each module has no version and resolves from the root pom's `dependencyManagement`, which already carries `jetbrains.annotations.version`.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
jimador: Same answer as the differ on #85: `dice-storage-autoconfigure` constructs it in `MetamodelAutoConfiguration`, and a host that wires the runner itself constructs it too. `internal` does not cross the module boundary.

dice-metamodel/src/main/kotlin/com/embabel/dice/metamodel/support/DefaultDriftCheckRunner.kt
jimador: Kept as is. A raw string would put the line breaks into the log message, and this stays a single log line. The concatenation is compile-time constant folding of two literals.

igordayen (approved):
@jimador - looks good, thank you!

PR #88, opened 2026-08-30
feat(dice-storage-autoconfigure): opt-in metamodel wiring

jimador:
PR 6, closing the metamodel train (#83 → #84 → #85 → #86 → #87 → #88). The governance loop becomes usable from a Spring Boot app, and it gets an operator. `MetamodelAutoConfiguration` in `dice-storage-autoconfigure` registers only when the host declares a `DeclaredSchemaSource` bean — declaring your schema is the opt-in — and then wires the loop: version store, drift-report store, observed-schema source, both differ roles, quarantine policy, `DriftSweepCapable`, drift runner, and the metamodel `SchemaCatalog`. Every wired collaborator is `@ConditionalOnMissingBean`. `embabel.dice.metamodel.enabled=false` removes the beans; `drift.mode` is `off` or `observe` (default); backend selection follows `embabel.dice.store.type`. Nothing in the wiring runs a check or moves a proposition: a check happens when the host calls `DriftCheckRunner.run()`, a quarantine when the host calls `DriftSweepCapable.sweep()`.

- **The operator surface.** `GovernanceOperationsService` in `dice` is the single entry point — read (latest reports, global and per context, plus the declaration in force and the version the last completed sweep reconciled against), run (a check answering the full impact a sweep would evaluate, `declaredDiff` included), release (one call per proposition, context-checked before it writes). `GovernanceController` serves the five operations on `/api/v1/metamodel`; `GovernanceTools` exposes the same five as `@LlmTool` tools; both call the one service. **EXPERIMENTAL**, marked `@ApiStatus.Experimental` in code across the stack. The routes activate only when the host writes `@Import(DiceRestConfiguration)` and a `DeclaredSchemaSource` is declared — the same activation as every other DICE controller, carried by a deferred import so the condition sees autoconfigured beans. `GovernanceTools` is host-constructed like `DiscoveryTools`; no auto-configuration registers a controller or a tool object. Per-report release is deliberately absent: nothing in the model ties a proposition to the report whose application held it, and the doc says what closing that needs.
- `ObserveOnlyDriftCheckRunner` is deleted and `DriftMode` reduces to `OFF | OBSERVE` — the runner is report-only structurally, so the decorator had nothing left to downgrade.
- The sweep bean announces status transitions to every `DiceEventListener` on the context; the derived-ownership schema wiring from #87 is registered here; the HTTP gate re-checks the declared schema directly.

**Breaking changes: none.** A host with no `DeclaredSchemaSource` bean sees no metamodel beans at all, proven by a context probe test.

```mermaid
flowchart TD
    A{DeclaredSchemaSource bean?} -- no --> Z[no metamodel beans at all]
    A -- yes --> B{metamodel.enabled}
    B -- false --> Z
    B -- true --> C[stores, observed schema, differs, policy, sweep, runner]
    C --> F[GovernanceOperationsService; REST via host Import of DiceRestConfiguration; tools host-constructed]
    F --> G[read reports and declared version]
    F --> H[run a check on demand]
    F --> I[release one proposition from quarantine]
    J[host calls DriftSweepCapable.sweep] --> K[quarantine, deliberate and evented]
```

## Landing

- Build the union of this stack's tip and the extraction-run stack's tip (#101) before either lands; the two share six recurring conflict files with recorded resolutions; `DiceStorageAutoConfiguration.kt` conflicts too and is resolved by hand from the union worktree's committed adaptation.
- Whichever stack lands second restacks and applies one adaptation: `ExtractionRunSchema` joins this stack's `DiceStorageSchema` derived-ownership contract (object declaration, `override` on `specs()`, bookkeeping relationship types, catalog and test beans). `DiceStorageSchemaRegistrationTest` fails until it is applied — that is the guard working.
- No consumer migration is required to land: the four-argument `undoSingleCollapse` stays as a deprecated overload (#92). Downstream callers move to `CollapseUndoCommand` when convenient.


dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,263 @@
...
+ *   `dryRun = false`.
+ * - `quarantine` — the real [DefaultDriftCheckRunner]. `run(dryRun = false)` moves stranded
+ *   propositions to `STALE` with a reason.
+ *
+ * The default is `observe`. Reporting is safe to leave running indefinitely; changing proposition
+ * state is a decision somebody makes on purpose. A mistyped type name in a declared schema costs
+ * five minutes under `observe` and marks every proposition mentioning that type stale under
+ * `quarantine`, so the escalation is opt-in in the direction where mistakes are recoverable.
+ *
+ * Nothing here schedules anything. This wires the capability to run a check; when a check runs is
```
jimador: What current path calls this runner? The stack wires the bean but provides no scheduler, endpoint, extraction hook, or downstream caller, so a running DICE app still catches nothing.

jimador: Added `MetamodelDriftSchedulingConfiguration`, the first real caller of `DriftCheckRunner.run` anywhere in DICE. It registers only when `embabel.dice.metamodel.drift.schedule` holds a cron expression. A scheduled tick always runs dry; quarantine still takes a deliberate call, and the KDoc records why a timer never moves propositions to STALE.

jimador:
The auto-configuration does not compose with the rest of DICE in four places:

- `DefaultDriftCheckRunner` is constructed without a `DiceEventListener`, so it uses `DEV_NULL`. Quarantine makes the canonical proposition `STALE` while `ProjectionLineageStaleCascade` never sees the transition and derived projection records remain active.
- Adding `DeclaredSchemaSource` under the default in-memory backend requires `PersistenceManager` because all three default governance collaborators are Drivine beans with no graph-store condition.
- A consumer `MetamodelDiffer` bean can be ignored. The runner factory accepts only `DeclaredObservedDiffer` and constructs a private `StructuralMetamodelDiffer` when that object does not also implement the other interface.
- Scheduling is conditioned only on any `DriftCheckRunner` plus a cron. A custom runner therefore still schedules when `embabel.dice.metamodel.enabled=false` or `drift.mode=off`.

Use the existing backend-selection and composite-event patterns, resolve both differ beans independently, and gate the scheduler with the same master enable/mode contract. The application-context tests should cover default in-memory startup, projection-lineage cascade, distinct custom differ beans, and a custom runner under both kill-switch settings.

jimador:
From the product side this is a governance loop with no operator.

With the defaults, `drift.mode=observe` decorates the runner so a live call is downgraded (`ObserveOnlyDriftCheckRunner.kt:46-54`), and the scheduled tick is always `dryRun=true` whatever the mode (`MetamodelAutoConfiguration.kt:318-341`). A dry run does not return the declared-vs-previous impact (thread on #86). An operator who sets the cron and reads the log learns that a check ran. There is no way to read a `DriftReport` outside a debugger, no way to trigger a live run, and no release. `DriftReportStore` has three bounded reads and nothing exposes them.

The positioning doc names operational tooling and memory inspection as the biggest gap against the field. This stack adds two new kinds of inspectable state and no way to inspect them. Before the stack is called shipped, define the minimal operator surface and put it in the last slice:

- read: latest drift reports, global and per context, plus the current declared version
- run: a drift check with an explicit `dryRun` flag that returns the same impact a live run would apply
- release: one call per proposition or per report that clears the quarantine

`DiscoveryController` and `DiscoveryTools` are the shape. REST plus agent tool over one service, the way `/why` is exposed. A CLI is the open question in `developer-docs-architecture.md:286-287` and can wait.

Second, `me` and `assistant` do not depend on `dice-metamodel` and declare no `DeclaredSchemaSource`, so the whole stack is inert for both consumers on landing. Both already build the `DataDictionary` this needs (`DocumentEntityExtraction.kt:39` and `KgEntityLookup.kt:66` in `me`). A small bean in `me` would be the first real declaration and would show whether `GovernedTypeSelector` and `SchemaAliases` fit a real dictionary. I would do that before adding any further metamodel slices.

Landing note. This stack and the extraction-run stack conflict on `DrivinePropositionRepository.kt` (this side is `4bf7e86` on #87), `CHANGELOG.md` and two integration tests. Restack whichever lands second and build the union once first.

jimador:
The loop has its operator. Landed as "Give the governance loop its operator surface", with the gate tightened in "Gate the governance HTTP surface on the declared schema directly".

**Read / run / release.** `GovernanceOperationsService` in `dice` is the single entry point: `latestReports` and `reportsInContext` read the drift log at both scopes (the store's three bounded reads now all have a caller), `currentDeclaredVersion` reports the declaration in force plus the version the last completed sweep reconciled against, `runCheck` answers the full impact a sweep would evaluate — including `declaredDiff` — and `releaseProposition` lifts one hold, checking the proposition's context before it writes. `GovernanceController` puts the five operations on `/api/v1/metamodel`; `GovernanceTools` exposes the same five as `@LlmTool` tools; both call the one service, on the `DiscoveryController`/`DiscoveryTools` shape. The HTTP half has its own auto-configuration and `.imports` entry, so a host that wants no endpoints excludes it by name in one line and keeps the service and the tools. CLI waits, as you said.

Two premises of the comment had already moved on the branch by the time it arrived: dry reports carry `declaredDiff` (the runner reads the swept baseline and feeds declared-vs-previous to the report), and `ObserveOnlyDriftCheckRunner` is deleted — the runner is report-only by construction ("Back live drift mode off to a deliberate sweep SPI", "Fix the autoconfiguration's composition and drop live-mode remnants"), so quarantine happens only when somebody calls `DriftSweepCapable.sweep` on purpose.

**Per-report release is deliberately absent.** Nothing in the model ties a quarantined proposition to the report whose application held it — a `DriftReport` has no identity, and the reason string carries schema names only, so every implementable semantic would also release holds a different check placed. It needs a `DriftReport` id plus a link written at quarantine time: its own slice. Per-proposition release ships now, and the doc and CHANGELOG state the gap.

**The `me` bean is recorded for the next wave**, ahead of any further metamodel slice — agreed that a real `DataDictionary` declaration is what proves `GovernedTypeSelector` and `SchemaAliases` fit.

**Landing note taken.** A merge-tree union of both stack tips was built green before either lands; whichever lands second restacks and applies the one recorded adaptation (the run schema joining the derived-ownership contract). The conflict resolutions for the shared files are recorded.

jimador:
Second pass. The operator surface, the sweep SPI and the two-value `DriftMode` are the right shape, and the class-level `@ConditionalOnBean(DeclaredSchemaSource)` gate holds: a host with no declared schema gets nothing. The remaining problems are all about what a host gets the moment it declares a schema. Declaring a schema opts a host into versioning. It must not opt them into HTTP endpoints or agent tools.

1. HTTP is on by default once a schema exists. `GovernanceHttpAutoConfiguration.kt:84-90` registers the controller on schema bean plus servlet plus `enabled` with `matchIfMissing = true`, through the `.imports` file. Every other DICE controller activates only through the host's explicit `@Import(DiceRestConfiguration)`. So a host that declares a `DeclaredSchemaSource` for versioning gets six routes under `/api/v1/metamodel`, two of them POSTs that run checks and release quarantines, with no auth, without ever importing DICE REST. That is what `RestEndpointSurfaceTest` in `me` caught on the union. Either add `GovernanceController` to `DiceRestConfiguration`'s import set and drop the autoconfiguration, or keep the autoconfiguration behind `embabel.dice.metamodel.http.enabled` defaulting to false. I would fold it into `DiceRestConfiguration` so there is one REST activation idiom.

2. `GovernanceTools` is registered as a bean. On main no DICE tool object is a bean. `DiscoveryTools` is constructed by the host, which is how a host chooses what an agent can call. A bean carrying `runDriftCheck` and `releaseProposition` reaches whatever discovers `@LlmTool` beans in the host. Keep it host-constructed like the others, or behind a property defaulting to false.

3. Nothing in this stack carries `@ApiStatus.Experimental`. Three hits across four modules, all on `PropositionStatus` constants, against fifty-one on the extraction stack. The CHANGELOG calls these surfaces experimental in prose. Mark the public types so a consumer sees it in the IDE. That is the whole signal for an M1 experimental surface.

4. Docs. `architecture.md:16` still says drift mode is `off / observe / quarantine`. `metamodel-wiring.md:233` lists the controller's gate without the schema bean and the enabled property. `metamodel-wiring.md` should also state that release is per proposition by design and that quarantine is applied only by a host call to `DriftSweepCapable.sweep`, since both are deliberate narrowings a reader will otherwise file as gaps.

The per-proposition release is the right call for now. The report has no identity, so per-report release would be guessing.

james-gsb:
All four on 3b505fe.

1. `GovernanceHttpAutoConfiguration` is deleted along with its `.imports` line. `GovernanceController` joins `DiceRestConfiguration`'s import set, so the six routes exist only after the host writes `@Import(DiceRestConfiguration)`, the same activation as every other DICE controller. One measured wrinkle: `@ConditionalOnBean` on a plain `@Import`ed class is answered before Spring Boot registers auto-configuration bean definitions, so it can never see an autoconfigured `GovernanceOperationsService`. The controller therefore arrives through a `DeferredImportSelector` at lowest precedence, which asks the condition after Boot's own selector has run. The host-facing idiom is unchanged. Wiring tests read live handler mappings across every import/loop combination; a mutation probe restoring the deleted autoconfiguration fails 9 of 16.

2. The `GovernanceTools` bean is gone. The class stays public and host-constructed; its KDoc shows the `DiscoveryTools` pattern, and a test asserts no autoconfigured context holds one.

3. `@ApiStatus.Experimental` across the stack: #83 (6c04af2), #85 (28031df), #86 (ec7b957), #87 (ac087aa), and this PR's service, controller, tools, DTOs, `MetamodelAutoConfiguration`, `MetamodelProperties` and `DriftMode`.

4. `architecture.md` names the two modes. `metamodel-wiring.md` documents the import activation, states release is per proposition by design, and that quarantine happens only through a host call to `DriftSweepCapable.sweep`.

Found while measuring: `DiscoveryController` and `PropositionPipelineController` gate on beans their auto-configurations supply, so on main those two may be silently absent exactly when the stores come from auto-configuration. Only the negative direction is tested today. Left untouched here; filed for its own fix with the same deferral.

jimador:
All four on 3b505fe.

1. `GovernanceHttpAutoConfiguration` is deleted along with its `.imports` line. `GovernanceController` joins `DiceRestConfiguration`'s import set, so the six routes exist only after the host writes `@Import(DiceRestConfiguration)`, the same activation as every other DICE controller. One measured wrinkle: `@ConditionalOnBean` on a plain `@Import`ed class is answered before Spring Boot registers auto-configuration bean definitions, so it can never see an autoconfigured `GovernanceOperationsService`. The controller therefore arrives through a `DeferredImportSelector` at lowest precedence, which asks the condition after Boot's own selector has run. The host-facing idiom is unchanged. Wiring tests read live handler mappings across every import/loop combination; a mutation probe restoring the deleted autoconfiguration fails 9 of 16.

2. The `GovernanceTools` bean is gone. The class stays public and host-constructed; its KDoc shows the `DiscoveryTools` pattern, and a test asserts no autoconfigured context holds one.

3. `@ApiStatus.Experimental` across the stack: #83 (6c04af2), #85 (28031df), #86 (ec7b957), #87 (ac087aa), and this PR's service, controller, tools, DTOs, `MetamodelAutoConfiguration`, `MetamodelProperties` and `DriftMode`.

4. `architecture.md` names the two modes. `metamodel-wiring.md` documents the import activation, states release is per proposition by design, and that quarantine happens only through a host call to `DriftSweepCapable.sweep`.

Found while measuring: `DiscoveryController` and `PropositionPipelineController` gate on beans their auto-configurations supply, so on main those two may be silently absent exactly when the stores come from auto-configuration. Only the negative direction is tested today. Left untouched here; filed for its own fix with the same deferral.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,414 @@
...
+ *
+ * `embabel.dice.metamodel.drift.mode` picks whether the runner bean is registered at all: `off` for
+ * stamps and history alone, `observe` (the default) for the check as well. See [DriftMode].
+ *
+ * ## Status transitions reach the application's listeners
+ *
+ * A sweep moves a proposition to `QUARANTINED`, and things downstream need to hear about it —
+ * `ProjectionLineageStaleCascade` marks the projection records derived from that proposition stale,
+ * and it can only do so if the transition is announced. So the sweep is built with every
+ * `DiceEventListener` bean the application registered, fanned out through a
```
jimador, line 94: `DefaultDriftCheckRunner` is constructed without a `DiceEventListener`, so it uses `DEV_NULL`. Quarantine makes the canonical proposition `STALE` while `ProjectionLineageStaleCascade` never sees the transition and derived projection records remain active.

jimador: The sweep bean announces status transitions to every `DiceEventListener` on the context through the composite-event pattern, so `ProjectionLineageStaleCascade` sees a quarantine the same way it sees any other transition. Covered by the projection-lineage cascade context test.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,414 @@
...
+ * when these conditions are evaluated, whichever order the two were declared in.
+ *
+ * The metamodel constraints ride along as a [SchemaCatalog] bean, the same way the proposition and
+ * lineage constraints do; Drivine's `SchemaManager` applies them idempotently on startup. The
+ * stores' MERGEs are only race-free under them, so they are required wherever the graph stores are.
+ */
+@ApiStatus.Experimental
+@AutoConfiguration(after = [DiceStorageAutoConfiguration::class])
+@ConditionalOnBean(DeclaredSchemaSource::class)
+@ConditionalOnProperty(
```
jimador, line 126: Adding `DeclaredSchemaSource` under the default in-memory backend requires `PersistenceManager` because all three default governance collaborators are Drivine beans with no graph-store condition. Use the existing backend-selection pattern.

jimador: Backend selection follows `embabel.dice.store.type`: the Drivine collaborators register under the graph condition, the in-memory ones otherwise. Default in-memory startup with a declared schema is a context test.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,414 @@
...
+    fun inMemoryMetamodelVersionStore(): SweptBaselineStore {
+        logger.debug("Wiring in-memory MetamodelVersionStore: InMemoryMetamodelVersionStore")
+        return InMemoryMetamodelVersionStore()
+    }
+
+    // ---- Pure-JVM collaborators, both backends ----
+
+    /**
+     * The shipped differ, registered under its concrete type so it resolves as both a
+     * [MetamodelDiffer] (two declarations compared) and a [DeclaredObservedDiffer] (a declaration
```
jimador, line 226: A consumer `MetamodelDiffer` bean can be ignored. The runner factory accepts only `DeclaredObservedDiffer` and constructs a private `StructuralMetamodelDiffer` when that object does not also implement the other interface. Resolve both differ beans independently.

jimador: Both differ roles are resolved independently, each `@ConditionalOnMissingBean`, so a host's `MetamodelDiffer` is honoured whether or not it also implements `DeclaredObservedDiffer`. Distinct custom differ beans are a context test.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,414 @@
...
+ * Requiring the bean puts the opt-in in the consumer's own code, where a reader can see it.
+ *
+ * `embabel.dice.metamodel.enabled=false` is a kill switch on top of that, for switching governance
+ * off in one environment while the bean stays in place. It is consulted only once the bean exists,
+ * so it changes nothing for an application that never declared a schema.
+ *
+ * ## Nothing here runs, and nothing here quarantines
+ *
+ * Every bean below is a capability sitting still. There is no scheduler and no startup hook: a check
+ * happens when the application calls [DriftCheckRunner.run], from a cron, an admin endpoint or a
```
jimador, line 78: Scheduling is conditioned only on any `DriftCheckRunner` plus a cron. A custom runner therefore still schedules when `embabel.dice.metamodel.enabled=false` or `drift.mode=off`. Gate the scheduler with the same master enable/mode contract.

jimador: The scheduler sits behind the same gate as the rest: `embabel.dice.metamodel.enabled` and `drift.mode`. A custom runner under either kill switch does not schedule; both settings are context tests.

dice/src/main/kotlin/com/embabel/dice/governance/GovernanceOperationsService.kt
```diff
@@ -0,0 +1,233 @@
...
+ * @param versionStore Where stamps live. Consulted to say whether the current declaration has ever
+ *   been recorded, and for the reconciled baseline when the store tracks one.
+ * @param driftReportStore The durable drift log.
+ * @param driftCheckRunner Performs a check.
+ * @param driftSweep The store-side quarantine operations. Only [DriftSweepCapable.releaseFromQuarantine]
+ *   is used here; nothing in this class sweeps.
+ * @param propositions Read to confirm a proposition's context before a release touches it.
+ */
+@ApiStatus.Experimental
+class GovernanceOperationsService(
```
jimador, line 79: From the product side this was a governance loop with no operator.

With the defaults, `drift.mode=observe` decorated the runner so a live call was downgraded, and the scheduled tick is always `dryRun=true` whatever the mode. An operator who sets the cron and reads the log learns that a check ran. There was no way to read a `DriftReport` outside a debugger, no way to trigger a live run, and no release. `DriftReportStore` has three bounded reads and nothing exposed them.

The positioning doc names operational tooling and memory inspection as the biggest gap against the field. This stack adds two new kinds of inspectable state and needs a way to inspect them. Before the stack is called shipped, define the minimal operator surface and put it in the last slice:

- read: latest drift reports, global and per context, plus the current declared version
- run: a drift check with an explicit `dryRun` flag that returns the same impact a live run would apply
- release: one call per proposition or per report that clears the quarantine

`DiscoveryController` and `DiscoveryTools` are the shape. REST plus agent tool over one service, the way `/why` is exposed. A CLI is an open question in the docs plan and can wait.

jimador: The loop has its operator. Landed as "Give the governance loop its operator surface", with the gate tightened in "Gate the governance HTTP surface on the declared schema directly".

**Read / run / release.** `GovernanceOperationsService` in `dice` is the single entry point: `latestReports` and `reportsInContext` read the drift log at both scopes (the store's three bounded reads now all have a caller), `currentDeclaredVersion` reports the declaration in force plus the version the last completed sweep reconciled against, `runCheck` answers the full impact a sweep would evaluate, `declaredDiff` included, and `releaseProposition` lifts one hold, checking the proposition's context before it writes. `GovernanceController` puts the five operations on `/api/v1/metamodel`; `GovernanceTools` exposes the same five as `@LlmTool` tools; both call the one service, on the `DiscoveryController`/`DiscoveryTools` shape. CLI waits.

Two premises of the comment had already moved on the branch by the time it arrived: dry reports carry `declaredDiff`, and `ObserveOnlyDriftCheckRunner` is deleted. The runner is report-only by construction, so quarantine happens only when somebody calls `DriftSweepCapable.sweep` on purpose.

**Per-report release is deliberately absent.** Nothing in the model ties a quarantined proposition to the report whose application held it. A `DriftReport` has no identity, and the reason string carries schema names only, so every implementable semantic would also release holds a different check placed. It needs a `DriftReport` id plus a link written at quarantine time: its own slice. Per-proposition release ships now, and the doc and CHANGELOG state the gap.

docs/design/metamodel-wiring.md
```diff
@@ -0,0 +1,308 @@
+# Metamodel wiring: turning governance on
+
+The governance pieces (stamps, diffs, drift checks, quarantine) are ordinary objects with
+constructors. This note covers how a Spring Boot application gets them without assembling them by
+hand, and what it has to do to switch them on.
+
+## The opt-in is a declared schema
+
+`MetamodelAutoConfiguration` activates when the application supplies a `DeclaredSchemaSource` bean,
```
jimador, line 9: The known consumers do not depend on `dice-metamodel` and declare no `DeclaredSchemaSource`, so the whole stack is inert for them on landing. Each already builds the `DataDictionary` this needs. A small `DeclaredSchemaSource` bean in one of them would be the first real declaration and would show whether `GovernedTypeSelector` and `SchemaAliases` fit a real dictionary. I would do that before adding any further metamodel slices.

jimador: **The host bean is recorded for the next wave**, ahead of any further metamodel slice. Agreed that a real `DataDictionary` declaration is what proves `GovernedTypeSelector` and `SchemaAliases` fit.

CHANGELOG.md
```diff
@@ -548,3 +548,83 @@ and the consumer PRs that deliver it).
     committed independently and stayed committed whatever happened next. Which exceptions trigger
     that rollback follows Spring's defaults: a `RuntimeException` or an `Error` rolls back and a
     checked exception commits, and custom rollback rules can override either behaviour.
+- Spring Boot auto-configuration for schema governance: `MetamodelAutoConfiguration` in
+  `dice-storage-autoconfigure`. It registers only when the application supplies a
```
jimador, line 551: Landing note. This stack and the extraction-run stack conflict on `DrivinePropositionRepository.kt` (this side is `4bf7e86` on #87), `CHANGELOG.md` and two integration tests. Restack whichever lands second and build the union once first.

jimador: **Landing note taken.** A merge-tree union of both stack tips was built green before either lands; whichever lands second restacks and applies the one recorded adaptation (the run schema joining the derived-ownership contract). The conflict resolutions for the shared files are recorded; the body's Landing section carries the procedure.

dice/src/main/kotlin/com/embabel/dice/web/rest/DiceRestConfiguration.kt
```diff
@@ -37,5 +40,38 @@ import org.springframework.context.annotation.Import
     PropositionPipelineController::class,
     MemoryController::class,
     DiscoveryController::class,
+    GovernanceControllerImport::class,
```
jimador, line 43: Declaring a schema opts a host into versioning. It must not opt them into HTTP endpoints. HTTP was on by default once a schema existed: `GovernanceHttpAutoConfiguration` registered the controller on schema bean plus servlet plus `enabled` with `matchIfMissing = true`, through the `.imports` file. Every other DICE controller activates only through the host's explicit `@Import(DiceRestConfiguration)`. So a host that declares a `DeclaredSchemaSource` for versioning got six routes under `/api/v1/metamodel`, two of them POSTs that run checks and release quarantines, with no auth, without ever importing DICE REST. That is what a host's endpoint-surface snapshot test caught on the union. Either add `GovernanceController` to `DiceRestConfiguration`'s import set and drop the autoconfiguration, or keep the autoconfiguration behind `embabel.dice.metamodel.http.enabled` defaulting to false. I would fold it into `DiceRestConfiguration` so there is one REST activation idiom.

jimador: `GovernanceHttpAutoConfiguration` is deleted along with its `.imports` line (3b505fe). `GovernanceController` joins `DiceRestConfiguration`'s import set, so the six routes exist only after the host writes `@Import(DiceRestConfiguration)`, the same activation as every other DICE controller. One measured wrinkle: `@ConditionalOnBean` on a plain `@Import`ed class is answered before Spring Boot registers auto-configuration bean definitions, so it can never see an autoconfigured `GovernanceOperationsService`. The controller therefore arrives through a `DeferredImportSelector` at lowest precedence, which asks the condition after Boot's own selector has run. The host-facing idiom is unchanged. Wiring tests read live handler mappings across every import/loop combination; a mutation probe restoring the deleted autoconfiguration fails 9 of 16.

Found while measuring: `DiscoveryController` and `PropositionPipelineController` gate on beans their auto-configurations supply, so on main those two may be silently absent exactly when the stores come from auto-configuration. Fixed with the same deferral in #102.

dice/src/main/kotlin/com/embabel/dice/agent/GovernanceTools.kt
```diff
@@ -0,0 +1,198 @@
...
+ * holding these tools can read and release in any context. A host decides whether an agent gets
+ * them at all by choosing to register them.
+ *
+ * Running a check writes a drift report and moves no proposition. Releasing a proposition is the one
+ * tool here that changes stored data, and it changes exactly one proposition.
+ *
+ * @param operations The single governance service every tool delegates to.
+ */
+@ApiStatus.Experimental
+class GovernanceTools(
```
jimador, line 69: `GovernanceTools` was registered as a bean. On main no DICE tool object is a bean. `DiscoveryTools` is constructed by the host, which is how a host chooses what an agent can call. A bean carrying `runDriftCheck` and `releaseProposition` reaches whatever discovers `@LlmTool` beans in the host. Keep it host-constructed like the others, or behind a property defaulting to false.

jimador: The `GovernanceTools` bean is gone. The class stays public and host-constructed; its KDoc shows the `DiscoveryTools` pattern, and a test asserts no autoconfigured context holds one.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelProperties.kt
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Whether the wiring registers a drift-check runner.
+ *
+ * There are two settings because there are two useful answers. Some applications want schema
+ * history and nothing else; most want the check as well. Neither setting can quarantine anything: a
+ * check reads, compares and writes a report, and moving a proposition is a separate call a host
+ * makes on `DriftSweepCapable` at a moment it picks.
+ */
+@ApiStatus.Experimental
```
jimador, line 29: Nothing in this stack carried `@ApiStatus.Experimental`. Three hits across four modules, all on `PropositionStatus` constants, against fifty-one on the extraction stack. The CHANGELOG calls these surfaces experimental in prose. Mark the public types so a consumer sees it in the IDE. That is the whole signal for an M1 experimental surface.

jimador: `@ApiStatus.Experimental` across the stack: #83 (6c04af2), #85 (28031df), #86 (ec7b957), #87 (ac087aa), and this PR's service, controller, tools, DTOs, `MetamodelAutoConfiguration`, `MetamodelProperties` and `DriftMode`.

docs/design/metamodel-wiring.md
```diff
@@ -0,0 +1,308 @@
...
+promise in the API that the stored data cannot keep.
+
+### What the wiring supplies, and what the host supplies
+
+`MetamodelAutoConfiguration` registers one thing here: the service.
+
+| Piece | Where it comes from |
+| --- | --- |
+| `GovernanceOperationsService` | the auto-configuration, when the governance loop is wired and a `DriftReportStore`, a `DriftCheckRunner`, a `DriftSweepCapable` and a `PropositionStore` are all on the context |
+| `GovernanceController` | `@Import(DiceRestConfiguration.class)` in the host's own configuration, when a `GovernanceOperationsService` is on the context |
```
jimador, line 240: Docs. `architecture.md` still said drift mode is `off / observe / quarantine`. `metamodel-wiring.md` listed the controller's gate without the schema bean and the enabled property. It should also state that release is per proposition by design and that quarantine is applied only by a host call to `DriftSweepCapable.sweep`, since both are deliberate narrowings a reader will otherwise file as gaps.

jimador: `architecture.md` names the two modes. `metamodel-wiring.md` documents the import activation, states release is per proposition by design, and that quarantine happens only through a host call to `DriftSweepCapable.sweep`.

dice-storage-autoconfigure/pom.xml
```diff
@@ -71,6 +81,17 @@
...
 
+        <!--
+          JetBrains annotations for @ApiStatus, declared the same way dice/pom.xml declares it.
+          It has to be declared here: another path reaches the same artifact at provided scope,
+          which wins the resolution and stops dice's own compile-scope copy travelling downstream.
+        -->
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>26.0.1</version>
```
igordayen: explicit versions slipped again...and ln.127

@jimador - candidate for parent pom? as it gets employed in few other poms

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
```diff
@@ -0,0 +1,414 @@
...
+     * The version store, returned as a [SweptBaselineStore] on purpose.
+     *
+     * Both shipped stores track the reconciled baseline, and a host needs that capability by name:
+     * after it has swept every context it meant to reconcile it calls
+     * [SweptBaselineStore.markSwept] itself. Spring predicts a bean's type from the factory method's
+     * return type, so declaring the narrower [MetamodelVersionStore] here would hide the baseline
+     * capability at injection time and leave a `SweptBaselineStore` injection point unresolvable.
+     */
+    @Bean
+    @ConditionalOnProperty(prefix = "embabel.dice.store", name = ["type"], havingValue = "graph")
```
igordayen: prefix repeatedly used, maybe use constant

igordayen:
From CODEX:

  1. dice/src/main/kotlin/com/embabel/dice/agent/GovernanceTools.kt:77

  The new governance tools are exposed with generic names like run_drift_check, latest_drift_reports, and
  release_quarantined_proposition.

  Example:

  - another module also exposes run_drift_check
  - a host registers both tool lists in one agent/MCP server
  - both tools now have the same public name

  At that point the tool registry has to either reject one, overwrite one, or leave selection ambiguous.

  So for this PR, the practical fix is to prefix each tool explicitly, for example:

  - embabel_dice_declared_schema_version
  - embabel_dice_latest_drift_reports
  - embabel_dice_run_drift_check

  If dotted tool names are valid in the target tool registry, embabel.dice.run_drift_check is clearer. If not, use underscores. The key point is that the prefix must reach Tool.Definition.name, not only metadata or docs.

There is also PR on ```embabel-agent``` that flags non-unique tool names and prefixes tool names with agent name,
see:
https://github.com/embabel/embabel-agent/pull/1833


  2. dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt:127

  The configuration prefixes are repeated directly in annotations:

  - embabel.dice.metamodel
  - embabel.dice.metamodel.drift
  - embabel.dice.store

  This is not a behavior bug, but it is easy to mistype later. Spring does not provide a class-level prefix that child
  @ConditionalOnProperty annotations inherit, so I would define constants instead:

  private const val DICE_PREFIX = "embabel.dice"
  private const val METAMODEL_PREFIX = "$DICE_PREFIX.metamodel"
  private const val DRIFT_PREFIX = "$METAMODEL_PREFIX.drift"
  private const val STORE_PREFIX = "$DICE_PREFIX.store"

  Then use those constants in @ConfigurationProperties and @ConditionalOnProperty. That keeps the public property names in one place.

  3. dice-storage-autoconfigure/src/test/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfigurationIntegrationTest.kt:163

  The comment says _DrivineSchema is currently reported as drift, but PR 87 already added INFRASTRUCTURE_LABELS and excludes  _DrivineSchema.

  So the comment is stale. It now tells the reader the opposite of the current behavior.

igordayen (commented):
@jimador - looks good, a few comments from my side and Codex, thank you

dice-storage-autoconfigure/pom.xml
jimador: Fixed in `cf6c0f3`. Both explicit versions are gone from this pom and from `dice/pom.xml`. `annotations` was already managed by the root pom; `mockito-kotlin` now is too, through a `mockito-kotlin.version` property and a `dependencyManagement` entry, since `embabel-agent-dependencies` does not manage it.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/MetamodelAutoConfiguration.kt
jimador: Fixed in `9e021e9`. The four prefixes live once in `DicePropertyPrefixes` and every annotation in `MetamodelAutoConfiguration` and `MetamodelProperties` reads from it. The collector and storage auto-configurations on `main` still spell theirs out; moving them onto the same object is a small follow-up outside this stack.

jimador:
On Codex items 2 and 3:

2. Fixed in `9e021e9`. The prefixes are `const val`s on one `internal object`, and the nine annotations across the two metamodel files reference it.
3. Fixed in `7edd5d7`. The comment now says `DrivineObservedSchemaSource` keeps `_DrivineSchema` out through `INFRASTRUCTURE_LABELS`, and the test asserts it.

Item 1 (tool name prefixes) is answered separately.

igordayen (approved):

PR #90, opened 2026-08-31
Source-revision provenance: contract and revision-aware storage

jimador:
PR 1 of the source-revision series (refs #64). Successive versions of one logical source become queryable without changing locator identity: stable key plus opaque revision, the S3/W3C-PROV pattern. `ProvenanceEntry` gains `sourceRevision`, an opaque provider-defined string participating in data-class equality, so evidence at two revisions over the same span survives a fold as two entries. `SourceLocator.key()` is untouched and revision-free — one document read at two revisions is still one source identity. `SourceRevisionRef(sourceKey, sourceRevision)` names one version of one source for queries. In `dice-storage`, `DERIVED_FROM` edges carry an `entryKey` identity and merge on it, so parallel revisions of one source persist as parallel edges; exact-text dedup unions novel evidence into the winner behind a structural source-identity guard; exact replays issue zero writes. Length ceilings on source key (2048) and revision (1024) are checked at construction, upstream of every hash and indexed write.

- **Revision queries moved onto a capability.** `SourceRevisionQueryCapable` carries the three context-scoped finders (`findBySourceKey`, `findBySourceRevision`, `findRevisionlessBySourceLocator`); `PropositionRepository` carries none of them. A shared default body over an ordinary context read would have answered an empty list on backends that store evidence without projecting it, and an empty list already means "nothing cites this source" — so a backend that cannot answer is absent from the type, and a caller probes with `as?`. `DrivinePropositionRepository` implements the capability with all three predicates pushed into Cypher; the in-memory and JSON stores answer through a provenance scan.
- **Source identity is bounded** via `SourceIdentityBounds`, with a rejection naming the limit it broke.

**Breaking changes: none** for source, JSON, or Java constructor descriptors — legacy revisionless callers keep byte-identical behavior, pinned by compat fixtures and, in the final PR of the series, an executed legacy-client jar. The stated boundary: Kotlin data-class synthetics (`copy`/`componentN`) on `ProvenanceEntry` move; Kotlin callers recompile. Stored graphs need no migration — an edge with no `sourceRevision` reads back as a revisionless entry.

**Opt-in and status:** **EXPERIMENTAL** (shape may change before 1.0). The trigger is a store implementing `SourceRevisionQueryCapable`; everything else is inert carriage.

**The identity rule the series rests on:**

```mermaid
flowchart LR
    L[SourceLocator] --> K["key(), revision-free and untouched"]
    K --> S[SourceRevisionRef.sourceKey]
    V[opaque sourceRevision] --> S
    S --> Q[exact revision queries via SourceRevisionQueryCapable]
    L --> P[ProvenanceEntry]
    V --> P
    P --> E[ProvenanceEvidenceKey, length-framed and injective]
    E --> D[DERIVED_FROM edges, one per evidence identity]
    D --> R[parallel revisions of one source persist as parallel edges]
```

**Series:** #91 storage proofs → #92 collector fold/undo → #93 entry points and REST. #64 stays open after the series lands; its bundle round-trip criterion gates on #46.

### DIFF
--- a/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
+++ b/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
@@ -31,6 +31,8 @@ import com.embabel.dice.proposition.GraphQueryCapable
 import com.embabel.dice.proposition.PropositionStatus
 import com.embabel.dice.proposition.PropositionStoreType
 import com.embabel.dice.provenance.ProvenanceEntry
+import com.embabel.dice.provenance.SourceLocator
+import com.embabel.dice.provenance.SourceRevisionRef
 import com.embabel.dice.query.graph.GraphNeighborhood
 import com.embabel.dice.query.graph.GraphPath
 import com.embabel.dice.query.graph.PropositionLineage
@@ -43,9 +45,53 @@ import org.drivine.query.QuerySpecification
 import org.drivine.query.dsl.*
 import org.slf4j.LoggerFactory
 import org.springframework.transaction.PlatformTransactionManager
+import org.springframework.transaction.annotation.Propagation
 import org.springframework.transaction.annotation.Transactional
+import org.springframework.transaction.support.TransactionSynchronizationManager
 import org.springframework.transaction.support.TransactionTemplate
 import java.time.Instant
+
+/**
+ * The three source-provenance queries, kept together so they stay in step with each other.
+ *
+ * Each scopes to the tenant first and then checks for a matching `DERIVED_FROM` edge, so provenance
+ * is only expanded for one tenant's propositions.
+ *
+ * There is deliberately no `USING INDEX` hint. Neo4j resolves hints at planning time and fails the
+ * whole query when the named index is absent, and `dice-storage`'s schema is adopter-supplied — the
+ * `(contextId, text)` constraint ships in `dice-storage-autoconfigure`'s catalog, so a host wiring
+ * this repository by hand may not have it. A hint would turn an optional dedup backstop into a hard
+ * runtime dependency for every source query. Without one, the planner still seeks on `contextId`
+ * where an index exists and falls back to a label scan where it does not.
+ */
+internal object SourceProvenanceQueryStatements {
+
+    val bySourceKey = """
+        MATCH (p:Proposition {contextId: ${'$'}contextId})
+        WHERE EXISTS {
+            MATCH (p)-[:DERIVED_FROM]->(:Source {key: ${'$'}sourceKey})
+        }
+        RETURN p.id AS id
+    """.trimIndent()
+
+    val bySourceRevision = """
+        MATCH (p:Proposition {contextId: ${'$'}contextId})
+        WHERE EXISTS {
+            MATCH (p)-[r:DERIVED_FROM]->(:Source {key: ${'$'}sourceKey})
+            WHERE r.sourceRevision = ${'$'}sourceRevision
+        }
+        RETURN p.id AS id
+    """.trimIndent()
+
+    val revisionlessBySourceKey = """
+        MATCH (p:Proposition {contextId: ${'$'}contextId})
+        WHERE EXISTS {
+            MATCH (p)-[r:DERIVED_FROM]->(:Source {key: ${'$'}sourceKey})
+            WHERE r.sourceRevision IS NULL
+        }
+        RETURN p.id AS id
+    """.trimIndent()
+}
 
 /**
  * Graph-backed [PropositionRepository] over Drivine / Neo4j.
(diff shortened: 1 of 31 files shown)


jimador:
The source-revision query contract is unsafe for a base repository SPI. The default methods can return an empty result on a backend that stores evidence but does not expose provenance through `findAll`. That turns an unsupported operation into a false negative.

Make this an explicit capability or a required primitive. A backend must either answer correctly or say it cannot.

There is also a scope leak in the graph projection: `Source` is merged globally by key, then its display value is overwritten by whichever context writes last. Source presentation must be scoped or immutable. Bound the externally supplied source and revision strings before hashing and indexed persistence.

dice/src/main/kotlin/com/embabel/dice/proposition/SourceRevisionQueryCapable.kt
```diff
@@ -0,0 +1,144 @@
...
+ * findByContextIdValue` on [PropositionStore]. Implementations live on the String variant, because
+ * [ContextId] is a value class: the typed method's JVM name is mangled and a Java backend cannot
+ * override it, so an implementation placed there would be invisible to Java callers. The String
+ * variant has an ordinary JVM signature and sits on the path of every typed call.
+ *
+ * A store whose reads already carry every provenance entry can pick up all three String variants
+ * from [ProvenanceScanningSourceRevisionQueries].
+ */
+@ApiStatus.Experimental
+interface SourceRevisionQueryCapable {
```
jimador, line 53: The source-revision query contract was unsafe for a base repository SPI. Default methods could return an empty result on a backend that stores evidence but does not expose provenance through `findAll`. That turns an unsupported operation into a false negative. Make this an explicit capability or a required primitive. A backend must either answer correctly or say it cannot.

jimador: Made an explicit capability. `SourceRevisionQueryCapable` carries the three context-scoped finders, with the String forms abstract, so a backend that implements it has promised to answer for its own storage and a backend that cannot answer is absent from the type; a caller probes with `as?`. `PropositionRepository` carries none of it. `DrivinePropositionRepository` pushes all three predicates into Cypher; the in-memory and JSON stores answer through a provenance scan.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -175,48 +255,260 @@ class DrivinePropositionRepository(
...
+    }
+
+    /**
+     * Upsert the shared `:Source` node and reject a key collision.
+     *
+     * Two structurally different locators that produce one key would otherwise take turns rewriting
+     * each other's node, so the incoming identity is compared against what is stored and a mismatch
+     * fails the write and leaves the shared node alone.
+     *
+     * Every property here is set once, on create — `display` included. The `:Source` node is global:
```
jimador: Scope leak in the graph projection: `Source` is merged globally by key, then its display value is overwritten by whichever context writes last. Source presentation must be scoped or immutable.

jimador: `:Source.display` is write-once: set on create only, so a later writer from another context leaves the label alone. Identity is unaffected; `display` never participated in `SourceLocator.key()`.

dice/src/main/kotlin/com/embabel/dice/provenance/SourceIdentityBounds.kt
```diff
@@ -0,0 +1,88 @@
...
+ * own hard ceiling on an indexed property that surfaces as an opaque database error deep inside a
+ * write. Checking here turns that into an [IllegalArgumentException] at the moment the value is
+ * offered, before anything is hashed and before any store is touched.
+ *
+ * The limits are deliberately roomy. Nothing a real connector produces comes close; they exist to
+ * stop a runaway or hostile value, and they are stated as constants so a caller can check its own
+ * input against the same number DICE will.
+ */
+@ApiStatus.Experimental
+object SourceIdentityBounds {
```
jimador, line 36: Bound the externally supplied source and revision strings before hashing and indexed persistence.

jimador: `SourceIdentityBounds` checks the source key (2048) and revision (1024) at construction, upstream of every hash and indexed write, with a rejection naming the limit it broke.

igordayen:
From Codex:

 - P1 dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt:403: :Source.display is still shared across contexts. The PR changed it to write-once, but that only changes “last writer wins” into “first writer wins.” Example:
    context A stores UriLocator("https://doc", display="A private label"); context B stores the same URI with display="B label". When B later reads its provenance, withRawProvenance() reads s.display from the shared :Source node, so B sees A’s label. Since display is presentation data, not identity, it should live on the DERIVED_FROM edge or be omitted from shared storage.

  - P2 dice/src/main/kotlin/com/embabel/dice/proposition/EventEmittingPropositionRepository.kt:68: the decorator always implements
    SourceRevisionQueryCapable, even when the delegate cannot answer source-revision queries. That weakens the contract documented on SourceRevisionQueryCapable: “probe with as?.” Example: a Spring component requiring SourceRevisionQueryCapable can receive an event-emitting wrapper around a plain repository, pass the type check, and only fail at call time. Prefer not exposing the capability unless the delegate supports it, or make the public contract say callers must check supportsSourceRevisionQueries before every use.

  - P2 dice-storage/src/main/kotlin/com/embabel/dice/storage/PropositionGraphMapper.kt:252: provenanceStorageEntryKey() length-frames    the full tuple, including chunkId and contentHash, but only source key and revision are bounded. Example: a connector can provide a    5 MB chunkId; this becomes part of entryKey and is also written as r.chunkId in DrivinePropositionRepository.kt:373-378. The new identity property should either hash the full tuple or bound every string field that participates in it.

jimador:
All three are right. Fixed in three commits on this branch, with the train above restacked on it.

- P1, `:Source.display`: in `90c22c0`. The label a writer supplies now travels on that writer's own `DERIVED_FROM` edge and provenance reads take it from there, with `coalesce` back to the node for rows written before the change. The node is still shared and its identity properties still write once; only the label moved. The integration test now has two contexts citing one URI and each reads back its own label.
- P2, decorator capability: in `d97c6b8`. `EventEmittingPropositionRepository` no longer declares `SourceRevisionQueryCapable`. A subclass carries it by delegation when the delegate has it, and `EventEmittingPropositionRepository.wrapping(delegate)` picks the shape, so `as?` on the wrapper answers the way it answers on the delegate.
- P2, entry-key bounds: in `4857b54`. `SourceIdentityBounds` gained `MAX_CHUNK_ID_LENGTH` (512) and `MAX_CONTENT_HASH_LENGTH` (256), checked in `ProvenanceEntry` next to the key check, so every string in the stored identity is bounded before anything is hashed or written.

igordayen (approved):

PR #91, opened 2026-08-31
Source-revision provenance: proven storage semantics

jimador:
**Breaking changes: none.** Additive tests, one public visibility widening (`ProvenanceEvidenceKey`), and query-plan work; no API signature changed.

PR 2 of the source-revision series (refs #64). This slice proves and pins what PR 1 built:
- **Query plans, captured and EXPLAINed**: the tenant-first finder path pins a `NodeIndexSeek` off the `contextId` range index (added to the test catalog; production already ships one), and a paired test proves the hint-free statements still plan and answer with no index at all — the old branch's `USING INDEX` hints are gone because they fail queries outright on adopter schemas without the constraint.
- **One codec**: the storage entry-key encoding delegates to the now-public domain `ProvenanceEvidenceKey`, with a golden literal locking the v1 wire format so a coordinated format edit can't ship green while orphaning stored keys. The unification changed stored bytes (the storage copy lacked the version prefix) — safe only because `entryKey` is unreleased within this same series, disclosed in the CHANGELOG.
- **Delete probed**: removing a proposition holding parallel revision edges of one source behaves correctly through the existing view cascade, with a bystander-safety test.
- **Validation topology**: one read-only preflight per distinct source key for fail-before-mutate ordering, with the atomic post-MERGE identity check inside `upsertSource` as the actual guarantee (TOCTOU closed; barrier-tested with two racing writers on Neo4j).

Tests: `dice` 1213, `dice-storage` 114. Review: three rounds; the TOCTOU, a mixed-batch validation bypass, and a plan-test flake (planner statistics) all caught, reproduced, fixed.

### DIFF
--- a/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
+++ b/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
@@ -464,7 +464,7 @@ class DrivinePropositionRepository(
                         r.startOffset = ${'$'}startOffset,
                         r.endOffset = ${'$'}endOffset,
                         r.contentHash = ${'$'}contentHash,
-                        r.display = ${'$'}sourceDisplay
+                        r.display = ${'$'}display
                     """.trimIndent()
                 ).bind(params),
             )
@@ -536,7 +536,7 @@ class DrivinePropositionRepository(
                   AND ((r.contentHash IS NULL AND ${'$'}contentHash IS NULL) OR r.contentHash = ${'$'}contentHash)
                 WITH r LIMIT 1
                 SET r.entryKey = ${'$'}entryKey,
-                    r.display = ${'$'}sourceDisplay
+                    r.display = ${'$'}display
                 RETURN count(r) AS adopted
                 """.trimIndent()
             ).bind(params).transform(Long::class.java)
@@ -593,6 +593,7 @@ class DrivinePropositionRepository(
         "sourceKey" to edge.source.key,
         "sourceKind" to edge.source.kind,
         "sourceDisplay" to edge.source.display,
+        "display" to edge.display,
         "sourceUri" to edge.source.uri,
         "sourcePath" to edge.source.path,
         "sourceContentHash" to edge.source.contentHash,
@@ -836,7 +837,8 @@ class DrivinePropositionRepository(
                     sourceRevision: r.sourceRevision,
                     sourceKey: s.key,
                     sourceKind: s.kind,
-                    sourceDisplay: coalesce(r.display, s.display),
+                    sourceDisplay: s.display,
+                    display: r.display,
                     sourceUri: s.uri,
                     sourcePath: s.path,
                     sourceContentHash: s.contentHash,
@@ -870,6 +872,7 @@ class DrivinePropositionRepository(
                 endOffset = (row["endOffset"] as? Number)?.toInt(),
                 contentHash = row["contentHash"] as? String,
                 sourceRevision = row["sourceRevision"] as? String,
+                display = row["display"] as? String,
             ),
         )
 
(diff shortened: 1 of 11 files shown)


jimador:
One commit added here while restacking on the #90 review fixes: `10527c5` hardens the plan-pinning fixture. After the display test on #90 was rewritten, `public source queries execute their production statements on tenant-first plans` failed on its first try in every full `dice-storage` run and passed on the rerun. Earlier tests in the class execute the same finder statements against a two-row store, where a scan and a seek cost the same, and Neo4j keeps that plan in its query cache until the replan interval passes. `db.resampleOutdatedIndexes` refreshes statistics but does not evict the plan, so the EXPLAIN answered with the two-row plan. The fixture now calls `db.prepareForReplanning`, which resamples every index and clears the cache. The full run is green on the first try with no flake.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
```diff
@@ -277,37 +283,60 @@ class DrivinePropositionRepository(
...
-        val source = PropositionGraphMapper.toDerivedFrom(entry).source
-        val stored = storedSourceIdentity(source.key) ?: return
-        require(sourceIdentityMatches(stored, source)) {
-            "Source key collision for '${source.key}': stored source identity differs from incoming locator"
-        }
+    private fun validateSourceIdentities(entries: List<ProvenanceEntry>) {
+        entries.map { PropositionGraphMapper.toDerivedFrom(it).source }
+            .groupBy(SourceNode::key)
+            .forEach { (key, sources) ->
+                val first = sources.first()
```
igordayen, line 342: sources. first() is hard to read . Could you please add a comment?

igordayen:
From CODEX:

Findings

  - P2 dice/src/main/kotlin/com/embabel/dice/provenance/ProvenanceEvidenceKey.kt:70: the public persisted key format uses String.length for frame lengths, but the contract only says “length-framed.” Example: "💾" has length 2 in Kotlin/JVM, 1 by Unicode code point count, and 4 in UTF-8 bytes. Since this codec is now public and stored on graph edges, another implementation can reasonably encode the same evidence differently. Either document/pin “UTF-16 code units” with a non-ASCII golden literal, or frame it by UTF-8 bytes for a language-neutral format.

  - P3 dice-storage/src/main/kotlin/com/embabel/dice/storage/model/DerivedFrom.kt:37: r.display is now written on the DERIVED_FROM edge, and raw reads prefer it (DrivinePropositionRepository.kt:419, :791), but the DerivedFrom relationship model has no display field.
    Example: if any view-based mapper path reads or writes provenance, it can only carry SourceNode.display, not the per-edge display,
    so it can lose the “writer’s own label” behavior the PR documents. If the view is truly delete-only, narrow/document that surface more aggressively; otherwise add display to DerivedFrom and pin it in mapper tests.

igordayen (commented):
@jimador - minor comment from my side and CODEX, thank you

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRepository.kt
jimador: Added in `4d220ac`. Every entry grouped under a key claims to describe the same source, so any one of them can stand for the group; the first is checked against the rest, then against what the graph already holds under that key. The comment now says exactly that above the line.

jimador:
Both taken.

**P2, frame length.** Frames now count the value's UTF-8 bytes (`a691fee`). `appendFrame` and the matcher both go through one `utf8Length`, the string comparison itself is unchanged, and the ASCII golden literal is byte-identical. A second golden literal pins the non-ASCII case: `uri:https://a/💾` frames as `18:` and the matcher rejects the same string framed as `16:` (UTF-16 units) or `15:` (code points). The KDoc, `docs/design/source-revisions.md`, and the CHANGELOG say which count it is.

**P3, `DerivedFrom.display`.** Added as a trailing optional field, so the Java arities `@JvmOverloads` already published still exist (`46365f2`). `toDerivedFrom` fills it from the locator, `toProvenanceEntry` prefers it and falls back to the node's label for edges written before the field existed, and the raw Cypher read now returns `r.display` and `s.display` as separate columns and lets the mapper do the fallback. `PropositionGraphMapperRevisionTest` pins all three cases.

igordayen (approved):
@jimador - looks good, approving, subject to rebasing, thanks

PR #92, opened 2026-08-31
Source-revision provenance: authoritative collector fold and undo

jimador:
PR 3 of the source-revision series (refs #64), stacked on #91. Collector fold and undo become authoritative under revisions. Fold refs record through the evidence codec, so a revisioned loser's evidence subtracts exactly on undo. Undo authorization is proven from durable facts the acting paths write themselves: the sweep records the survivor it actually merged into (`CollectorRecord.mergedIntoId`, written after the applied save), and a completed undo stamps `undoneAt` before the member's restore, so every crash window is retry-recoverable and a stranded restore resumes without touching evidence. The dry-run trace contract holds: previews record everything, gated by the run header, and can never authorize.

- **Undo fails closed in every direction the review named.** `undoSingleCollapse` takes a `CollapseUndoCommand(contextId, survivorId, retiredId)`, and both propositions must live in that context — a mismatch throws before any write, so ids found anywhere can no longer reverse a collapse in a context the caller has no standing in. A `CollectorRecordStore` is required; the status-only fallback is gone, so nothing skips the "did the collector apply this merge" question. Authorization needs a live run header and a `TRANSITIONED` record naming this survivor.
- **Evidence subtraction is an atomic capability.** `PropositionStore.subtractProvenance` and its read-modify-write default are removed; `ProvenanceSubtractionCapable` states the contract — read and write land as one step, evidence added concurrently survives, subtracting from a deleted proposition answers null and writes nothing. Undo requires the capability and refuses when the store cannot answer. A 200-round race test proves the atomicity; the replaced shape loses the concurrent addition on round one.
- **A deleted survivor ends the undo.** When subtraction answers null, the undo is abandoned with a warn log, nothing is written, and the member stays retired — saving the stale in-memory copy would have recreated another writer's deleted proposition, folded evidence and all. The deletion wins; the KDoc says how late a deletion can land per store.
- `subtractProvenance` semantics pinned on both store paths; connector keys pinned injective where collisions once lived.

**Breaking changes: none for callers of `undoSingleCollapse`; one source-level break for a store that overrode `subtractProvenance`.** Both have a mechanical migration.

*`undoSingleCollapse`'s parameter list.* The four-argument form stays as a `@Deprecated` overload with the body it shipped with — it checks neither context ownership nor the audit records — so an unmigrated caller keeps the behavior it had and gets a compiler warning pointing at the command form. Removed in the next minor release. The guarded form:

```kotlin
// before
undoSingleCollapse(traceQuery, propositionStore, survivorId, retiredId)
// after
undoSingleCollapse(
    command = CollapseUndoCommand(contextId, survivorId, retiredId),
    traceQuery = traceQuery,
    propositions = propositionStore,   // must be ProvenanceSubtractionCapable
    collectorRecords = collectorRecordStore,
)
```

A caller with no `ContextId` has to obtain one, which is the point. Downstream callers migrate with this snippet when convenient; their own hand-rolled ownership checks become redundant once they do.

*`PropositionStore.subtractProvenance`.* A store that overrode it declares `ProvenanceSubtractionCapable` and drops the `override`; `DrivinePropositionRepository` is adapted in this change — its one-statement edge delete already satisfied the contract, so only the supertype list moved. A store that never mentioned the operation compiles unchanged.

No stored data changes. `CollectorRecord`'s two new fields ride `@JvmOverloads` and decode absent-as-null for every stored row.

```mermaid
sequenceDiagram
    participant U as undoCollapse
    participant P as ProvenanceSubtractionCapable
    U->>P: subtractProvenance(survivor, refs)
    alt survivor already deleted
        P-->>U: null
        Note over U: abandon, warn, write nothing
        Note over U: member stays retired
    else subtraction applied
        P-->>U: survivor minus this member's evidence
        U->>P: save survivor, final state, event-carrying
        U->>P: stamp undoneAt on the collector record
        U->>P: restore member to its prior status
    end
```

The stamp sits between the evidence writes and the restore, which is what makes every crash window recoverable; the crash matrix is in the KDoc.

### DIFF
--- a/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
+++ b/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
@@ -44,7 +44,8 @@ import java.time.Instant
  * - `(:CollectorDecision {id, runId, contextId, componentId, survivorId, action, createdAt})`
  *   with `id = "runId|componentId"`, plus one child
  *   `(:CollectorRetired {id, runId, contextId, propositionId, priorStatus, foldedGrounding,
- *   foldedProvenanceRefs, foldedSourceIds})-[:RETIRED_IN]->(:CollectorDecision)` per retired
+ *   foldedProvenanceRefs, foldedSourceIds, foldedProvenanceEvidenceKeys})
+ *   -[:RETIRED_IN]->(:CollectorDecision)` per retired
  *   proposition, so a reversal has everything a merging sweep folded onto the survivor.
  *
  * Every write is a single `UNWIND $rows AS r ...` round trip (see [com.embabel.dice.storage.CollectorTraceRowMappers] for
@@ -227,7 +228,8 @@ class DrivineCollectorTraceStore(
                 retired: [r IN retiredNodes WHERE r IS NOT NULL | {
                     propositionId: r.propositionId, priorStatus: r.priorStatus,
                     foldedGrounding: r.foldedGrounding, foldedProvenanceRefs: r.foldedProvenanceRefs,
-                    foldedSourceIds: r.foldedSourceIds
+                    foldedSourceIds: r.foldedSourceIds,
+                    foldedProvenanceEvidenceKeys: r.foldedProvenanceEvidenceKeys
                 }]
             } AS row
             """.trimIndent(),
--- a/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
+++ b/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
@@ -270,6 +270,22 @@ class DrivineCollectorTraceStore(
         }.onFailure { logger.warn("Skipping unreadable CollectorDecision row: {}", it.message) }.getOrNull()
     }
 
+    /** The newest decision that retired [propositionId]. Never one where it only survived. */
+    @Transactional(readOnly = true)
+    override fun findDecisionRetiring(propositionId: String): CollectorDecision? {
+        val node = queryRows(
+            """
+            MATCH (:CollectorRetired {propositionId: ${'$'}propositionId})-[:RETIRED_IN]->(d:CollectorDecision)
+            RETURN d ORDER BY d.createdAt DESC LIMIT 1
+            """.trimIndent(),
+            mapOf("propositionId" to propositionId),
+        ).firstOrNull() ?: return null
+        return runCatching {
+            val decisionId = node["id"]?.toString().orEmpty()
+            CollectorDecisionRowMapper.fromRow(node, retiredFor(decisionId))
+        }.onFailure { logger.warn("Skipping unreadable CollectorDecision row: {}", it.message) }.getOrNull()
+    }
+
     private fun retiredFor(decisionId: String): List<RetiredProposition> =
         queryRows(
             "MATCH (ret:CollectorRetired {decisionId: \$decisionId}) RETURN ret",
--- a/dice/src/main/kotlin/com/embabel/dice/proposition/EventEmittingPropositionRepository.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/proposition/EventEmittingPropositionRepository.kt
@@ -49,10 +49,10 @@ import org.slf4j.LoggerFactory
  * Throw isolation is the listener's responsibility — wrap the listener in `SafeDiceEventListener`
  * if you need graceful degradation.
  *
- * This class does not itself implement `SourceRevisionQueryCapable`, so it carries the capability
- * only when its delegate does: use [wrapping] to build the right shape for whatever delegate you
- * hand it, so a caller's `as? SourceRevisionQueryCapable` probe on the wrapper answers the same way
- * it would on the delegate.
+ * This class carries capabilities only when its delegate does. The [wrapping] factory method
+ * picks the right shape for whatever delegate you hand it, so a caller's `as? SourceRevisionQueryCapable`
+ * or `as? ProvenanceSubtractionCapable` probe on the wrapper answers the same way it would on the
+ * delegate.
  *
  * Example usage:
  * ```kotlin
@@ -65,37 +65,37 @@ import org.slf4j.LoggerFactory
  * @property delegate The underlying repository. All non-write methods forward here.
  * @property listener Notified after each persist. Defaults to [DiceEventListener.DEV_NULL] (no-op).
  */
-open class EventEmittingPropositionRepository(
-    protected val delegate: PropositionRepository,
+open class EventEmittingPropositionRepository<D : PropositionRepository>(
+    protected val delegate: D,
     private val listener: DiceEventListener = DiceEventListener.DEV_NULL,
-) : PropositionRepository by delegate, ProvenanceSubtractionCapable {
+) : PropositionRepository by delegate {
 
     private val logger = LoggerFactory.getLogger(EventEmittingPropositionRepository::class.java)
 
     companion object {
         /**
          * Build the decorator over [delegate], picking the shape that matches what [delegate] can
-         * do: when it implements [SourceRevisionQueryCapable], the returned instance does too and
-         * forwards to it, so a caller's `as?` probe on the wrapper answers the way it answers on the
-         * delegate. Over a plain repository, the returned instance carries no source-revision
-         * surface at all.
+         * do: when it implements [SourceRevisionQueryCapable] and [ProvenanceSubtractionCapable],
+         * the returned instance does too and forwards to both, so a caller's `as?` probe on the
+         * wrapper answers the way it answers on the delegate. Over a delegate with only provenance
+         * subtraction, the wrapper carries that capability alone. Over a plain repository, the
+         * returned instance carries no source-revision or provenance-subtraction surface at all.
          */
         @JvmStatic
         @JvmOverloads
         fun wrapping(
             delegate: PropositionRepository,
             listener: DiceEventListener = DiceEventListener.DEV_NULL,
-        ): EventEmittingPropositionRepository =
-            if (delegate is SourceRevisionQueryCapable) {
-                SourceRevisionEventEmittingPropositionRepository(delegate, listener)
-            } else {
-                EventEmittingPropositionRepository(delegate, listener)
+        ): EventEmittingPropositionRepository<out PropositionRepository> =
+            when {
+                delegate is SourceRevisionQueryCapable && delegate is ProvenanceSubtractionCapable ->
+                    SourceRevisionEventEmittingPropositionRepository(delegate, listener)
+                delegate is ProvenanceSubtractionCapable ->
+                    ProvenanceSubtractingEventEmittingPropositionRepository(delegate, listener)
+                else ->
+                    EventEmittingPropositionRepository(delegate, listener)
             }
     }
-
-    /** The delegate's atomic evidence subtraction, if it has one. Resolved once at construction. */
-    private val delegateSubtraction: ProvenanceSubtractionCapable? =
-        delegate as? ProvenanceSubtractionCapable
 
     /**
      * Persists via the delegate, then emits one lifecycle event carrying the saved instance.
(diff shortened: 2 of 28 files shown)


dice/src/main/kotlin/com/embabel/dice/spi/CollectorSignals.kt
```diff
@@ -165,55 +187,368 @@ data class CollapseUndoResult(
...
 
-    val survivor = propositions.findById(survivorId) ?: return null
+    val refsToSubtract = retirement.provenanceRefsForUndo().filterNot { it in stillNeededProvenanceRefs }
+
+    // Evidence goes first and by name. subtractProvenance deletes exactly these refs against
+    // whatever the store holds now, so evidence another writer adds while this undo is running is
+    // left alone — naming what stays instead would replace it away. The save then carries grounding
+    // and source ids over the survivor as the subtraction left it, and goes last because save is
+    // the write a decorator instruments: the event a listener receives describes the final state.
+    val subtracted = propositions.subtractProvenance(survivorId, refsToSubtract) ?: survivor
```
jimador: `subtractProvenance` returns null when the survivor was deleted after the earlier read. Falling back to the stale object makes the next save recreate another writer's deleted proposition, so this should return without saving.

jimador: Returns without saving now. A null answer means the survivor was deleted after this function read it, so the undo is abandoned with a warn log and the member stays retired. The deletion wins.

jimador:
Undo currently fails open in three independent ways:

- The command takes proposition IDs without a `ContextId`, so it can mutate another context if those IDs are supplied.
- A missing `CollectorRecordStore` permits undo. A dry-run record can also become authorization for a later unrelated status change.
- The default `subtractProvenance` implementation is a non-atomic read/replace sequence, so concurrent evidence can be lost and a deleted proposition can be recreated.

Require the context in the command, require an auditable collector record for destructive undo, and make provenance subtraction an explicit atomic store capability. Add cross-context and concurrent-write tests.

dice/src/main/kotlin/com/embabel/dice/spi/CollectorSignals.kt
```diff
@@ -159,27 +182,332 @@ data class CollapseUndoResult(
...
+ * context the caller has no business writing to, and every deployment that cared about that had to
+ * bolt its own ownership check on in front. [undoSingleCollapse] now owns the check: both
+ * propositions have to live in [contextId], and an id from elsewhere is refused with
+ * [CollapseUndoContextMismatchException] before anything is written.
+ *
+ * @property contextId the context both propositions must belong to
+ * @property survivorId the collapse's survivor — must match the decision that retired [retiredId]
+ * @property retiredId the one retired member to restore
+ */
+data class CollapseUndoCommand(
```
jimador, line 219: Undo failed open: the command took proposition ids without a `ContextId`, so it could mutate another context if those ids were supplied. Require the context in the command and add cross-context tests.

jimador: `undoSingleCollapse` takes a `CollapseUndoCommand(contextId, survivorId, retiredId)`, and both propositions must live in that context; a mismatch throws `CollapseUndoContextMismatchException` before any write, ahead of the survivor-mismatch check so a caller from another context learns only that it was refused. Cross-context, mixed-context and probing-shape tests are in `CollectorUndoCapabilityTest`. The four-argument form stays as a `@Deprecated` overload for one release.

dice/src/main/kotlin/com/embabel/dice/spi/CollectorSignals.kt
```diff
@@ -159,27 +182,332 @@ data class CollapseUndoResult(
...
+    "Proposition $propositionId belongs to context ${actualContextId.value}, and the undo was " +
+        "issued for context ${commandedContextId.value}; refusing to touch it",
+)
+
+/**
+ * The undo was wired with something it cannot work from: no audit records to authorize it, or a
+ * proposition store that cannot subtract evidence atomically. Both are configuration mistakes, and
+ * both are thrown before the undo reads or writes anything.
+ */
+class CollapseUndoConfigurationException(message: String) : IllegalStateException(message)
```
jimador, line 253: A missing `CollectorRecordStore` permitted undo. A dry-run record could also become authorization for a later unrelated status change. Require an auditable collector record for destructive undo.

jimador: A `CollectorRecordStore` is required; the status-only fallback is gone. Authorization needs a live run header and a `TRANSITIONED` record whose `mergedIntoId` names this survivor; a dry-run record and an unrelated transition both refuse, and a completed undo stamps `undoneAt` before the restore so every crash window is retry-recoverable.

dice/src/main/kotlin/com/embabel/dice/proposition/ProvenanceSubtractionCapable.kt
```diff
@@ -0,0 +1,75 @@
...
+ *   [com.embabel.dice.provenance.ProvenanceEvidenceKey] names one entry, and a bare locator key
+ *   from before revisions existed reaches revisionless entries for that source only. Every entry
+ *   the refs do not name survives untouched.
+ * - **No resurrection.** Subtracting the last entry from a proposition another writer has already
+ *   deleted answers null and writes nothing. The proposition stays deleted.
+ *
+ * A decorator wrapping a capable store carries this type and forwards to its delegate;
+ * [EventEmittingPropositionRepository] is the worked example.
+ */
+interface ProvenanceSubtractionCapable {
```
jimador, line 61: The default `subtractProvenance` implementation was a non-atomic read/replace sequence, so concurrent evidence could be lost and a deleted proposition recreated. Make provenance subtraction an explicit atomic store capability and add concurrent-write tests.

jimador: `PropositionStore.subtractProvenance` and its read-modify-write default are removed; `ProvenanceSubtractionCapable` states the contract: read and write land as one step, evidence added concurrently survives, subtracting from a deleted proposition answers null and writes nothing. Undo requires the capability and refuses when the store cannot answer. A 200-round race test proves the atomicity; the replaced shape loses the concurrent addition on round one.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
```diff
@@ -44,7 +44,8 @@ import java.time.Instant
  * - `(:CollectorDecision {id, runId, contextId, componentId, survivorId, action, createdAt})`
  *   with `id = "runId|componentId"`, plus one child
  *   `(:CollectorRetired {id, runId, contextId, propositionId, priorStatus, foldedGrounding,
- *   foldedProvenanceRefs, foldedSourceIds})-[:RETIRED_IN]->(:CollectorDecision)` per retired
+ *   foldedProvenanceRefs, foldedSourceIds, foldedProvenanceEvidenceKeys})
```
igordayen, line 47: packase dice/storage includes generic and Drivine-specific artifacts. Consider moving Drivine artifacts into sub-package

dice/src/main/kotlin/com/embabel/dice/proposition/EventEmittingPropositionRepository.kt
```diff
@@ -171,6 +175,28 @@ open class EventEmittingPropositionRepository(
...
+     *
+     * This decorator instruments [save] alone, so the subtraction passes through unannounced —
+     * the same treatment the other provenance operations get through `by delegate`. Carrying the
+     * capability type matters because Kotlin's interface delegation only covers
+     * [PropositionRepository]: without this, wrapping a capable store would hide the capability
+     * from every caller that probes for it, and collector undo would refuse.
+     */
+    override fun subtractProvenance(propositionId: String, provenanceRefs: List<String>): Proposition? =
+        (
+            delegateSubtraction
```
igordayen: kotlin "required" ?(it will be as far as I remember, IllegalArgumentException)

dice/src/main/kotlin/com/embabel/dice/proposition/ProvenanceSubtractionCapable.kt
igordayen, line 61: very nice, in line with "agent" Capability model

igordayen:
FROM CODEX:

 - P2 dice/src/main/kotlin/com/embabel/dice/spi/CollectorSignals.kt:459: the full undo can still lose concurrently added evidence on
    stores whose save replaces provenance. Example with InMemoryPropositionRepository: survivor has [A, B], undo subtracts B and gets
    stale [A]; another extraction adds C, store now has [A, C]; then undo calls save(subtracted...) at CollectorSignals.kt:468, and
    InMemoryPropositionRepository.save replaces the row at InMemoryPropositionRepository.kt:65, putting it back to [A]. That violates
    the PR’s “evidence another extraction adds while this undo is running survives it” claim. The subtraction itself is atomic, but the
    following survivor save reintroduces a read/write window for replacing stores.

  - P2 dice/src/main/kotlin/com/embabel/dice/spi/CollectorSignals.kt:400: undo looks up
    traceQuery.findDecisionForProposition(retiredId), but that API may return a decision where the same id was the survivor, not the
    retired member. Concrete case: run 1 folds A into B, so B is a survivor in the trace; run 2 folds B into C; undoing B -> C can
    fetch run 1 first, fail to find B in decision.retired at line 401, and return null even though run 2 records authorize the undo.
    This is deterministic in DrivineCollectorTraceStore, which checks survivor matches before retired matches at
    DrivineCollectorTraceStore.kt:259-266, and possible in memory due arbitrary map iteration at InMemoryCollectorTraceStore.kt:83-86.
    Undo needs a retired-member-specific decision lookup, ideally scoped by the authorizing run.

jimador:
Both are real and both are fixed on this branch.

**Retired-member lookup.** `CollectorTraceQuery` gains `findDecisionRetiring(propositionId)`: the newest decision in which the id was a retired member, never one it survived. Both undo overloads and the `findRetirement` default use it. Drivine orders by `d.createdAt` and takes one row; the in-memory store keeps recording order. The chained case is a test on both stores: A folded into B in run 1, B into C in run 2, and undoing B out of C now finds run 2. The "shadowed decision" paragraph in the KDoc is rewritten to match.

**The survivor save.** Gone. `ProvenanceSubtractionCapable` now has `subtractFoldedEvidence(propositionId, provenanceRefs, grounding, sourceIds)`, which takes all three parts of a fold off in one atomic step; `subtractProvenance` is the same call with empty grounding and source ids. In memory that is one `compute`, on Drivine one statement that deletes the named edges, prunes orphans, and rewrites the two list properties together. The guarded undo calls it once and never saves the survivor, so a replacing `save` cannot write the evidence back from the undo's copy. Tests pin that no `save` reaches the survivor and that evidence another writer adds after the subtraction is still there afterwards. The crash matrix in the KDoc and the design doc now has three steps. Both new SPI members are abstract on experimental interfaces in an unreleased train; the CHANGELOG says so.

igordayen (approved):
@jimador  looks good, have minor comments on packaging, thx

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineCollectorTraceStore.kt
jimador: Agreed. Moving the Drivine types under their own sub-package touches most of `dice-storage`, and with the train open above this PR every rename would ripple through each rebase. It goes in a follow-up right after the train lands, as one mechanical commit.

dice/src/main/kotlin/com/embabel/dice/proposition/EventEmittingPropositionRepository.kt
jimador: The throw is gone, not reworded, in c3963696. `require` would have turned a missing backend capability into an `IllegalArgumentException` at call time, which is still a runtime answer to a question the type should settle. The decorator is now generic over its delegate and only carries `ProvenanceSubtractionCapable` when the delegate does, through `ProvenanceSubtractingEventEmittingPropositionRepository`, with `SourceRevisionEventEmittingPropositionRepository` for a store that also answers source-revision queries. `wrapping` picks the shape, so an `as?` probe on the wrapper answers the way it does on the delegate, and there is no cast and nothing to throw inside the classes. Changelog carries the compatibility note.

PR #93, opened 2026-08-31
Source-revision provenance: entry points and REST

jimador:
PR 4, closing the source-revision series (refs #64), stacked on #92. Revisions reach the entry points, the pipeline stamp, and REST. `SourceAnalysisContext.sourceRevision` is the only channel a revision travels on, guarded at construction: a revision requires a locator whose key it matches, and every `copy` re-checks. Both the direct entry points and the async event path feed one `buildContext`, so the two paths carry a revision identically, and the pipeline stamps it into each `ProvenanceEntry`. On REST, `POST /extract` and `POST /extract/file` take optional `sourceLocator` and `sourceRevision`; malformed combinations are rejected with 400 before the pipeline runs, and over-long identity values are answered at the edge with the limit named. A revisioned request derives context-scoped, chunker-stable chunk ids, so a replay grounds onto its own rows and two tenants can never share one. `/why` gains a `provenance` array serialized in stable evidence-key order across all backends.

- **`POST /extract` answers with what the store holds.** `save` is the authority: exact-text dedup hands back the existing canonical row under its own id, and a merge or reinforcement writes the revised proposition while the freshly extracted one is never stored. Both cases used to leave the response naming an id no read could resolve. Every proposition id in an extract response now resolves through `findById`, asserted over the dedup, merge, contradiction and reinforcement paths.
- A colon inside a `connectorId` is accepted — `ConnectorRef` escapes its own connector id injectively, so the guard that refused it is gone and `gmail:eu-west` round-trips unharmed.
- Bounds are answered at the request edge with the broken limit and offending length in the response body.

**Breaking changes: none** for source, JSON, or Java constructor descriptors — proven at runtime by a legacy client jar compiled against pre-revision `main`, executed against this tree with negative controls. Kotlin data-class synthetics stay the stated boundary. **One behavioral fix to call out:** the extract response's content changes as above — same wire shape, same fields and arity, and an extraction creating fresh propositions answers exactly as before, while merged and deduplicated extractions now report canonical stored ids. A `/why` consumer that rejects unknown JSON properties needs to allow the new `provenance` field.

**Opt-in and status:** **EXPERIMENTAL** like the rest of the series; requests carrying no locator and no revision behave byte-identically to before.

Consumer smoke: a host application's 727-test DICE subset is identical against this series and unmodified `main`; the one compile break seen was snapshot skew in an unrelated starter, isolation-confirmed. #64 stays open after the series lands — its bundle round-trip criterion gates on #46.


jimador:
The REST endpoint discards the canonical proposition returned by `save` and builds its response from the pre-save extraction result. When persistence deduplicates, the response can advertise an ID that does not exist.

Use the saved canonical proposition for both the response and downstream work.

The connector validation also rejects `:`, although the domain type already defines escaping for colon-bearing identities. The transport should preserve the domain contract. Add explicit length limits for the source identity and revision before they reach hashing and indexed storage.

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
```diff
@@ -91,21 +104,43 @@ class PropositionPipelineController(
...
+        )
         val result = propositionPipeline.processChunk(chunk, context)
 
         // Persist what revision says to keep — both the freshly extracted propositions and any
         // revised originals (e.g. an existing proposition retired to CONTRADICTED), not just the new ones.
-        result.propositionsToPersist().forEach { proposition ->
-            propositionRepository.save(proposition)
+        //
+        // save() is the store's answer about what it now holds, and that can be a different row from
+        // the one offered: exact-text dedup hands back the existing canonical proposition under its
```
jimador, line 136: The REST endpoint discarded the canonical proposition returned by `save` and built its response from the pre-save extraction result. When persistence deduplicates, the response can advertise an id that does not exist. Use the saved canonical proposition for both the response and downstream work.

jimador: `POST /extract` answers with what the store holds. `save` is the authority: exact-text dedup hands back the existing canonical row under its own id, and a merge or reinforcement writes the revised proposition while the freshly extracted one is never stored. Every proposition id in an extract response now resolves through `findById`, asserted over the dedup, merge, contradiction and reinforcement paths.

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
```diff
@@ -330,4 +480,88 @@ class PropositionPipelineController(
...
+     */
+    @ExceptionHandler(InvalidSourceProvenanceException::class)
+    fun handleInvalidSourceProvenance(
+        e: InvalidSourceProvenanceException,
+    ): ResponseEntity<ExtractErrorResponse> =
+        ResponseEntity.badRequest().body(ExtractErrorResponse(e.message ?: DEFAULT_REJECTION_REASON))
+
+    /**
+     * Builds the locator the request names, rejecting any field combination that would silently
+     * mean something else. `connectorId` belongs to `connector` locators alone; the other three
```
jimador, line 520: The connector validation rejected `:`, although the domain type already defines escaping for colon-bearing identities. The transport should preserve the domain contract.

jimador: A colon inside a `connectorId` is accepted. `ConnectorRef` escapes its own connector id injectively, so the guard that refused it is gone and `gmail:eu-west` round-trips unharmed.

dice/src/main/kotlin/com/embabel/dice/web/rest/PropositionPipelineController.kt
```diff
@@ -238,10 +295,83 @@ class PropositionPipelineController(
...
 
+    private fun parseSourceLocator(json: String?): SourceLocatorInputDto? =
+        if (json.isNullOrBlank()) null else objectMapper.readValue(json)
+
+    /**
+     * Turns the request's locator and revision fields into the typed pair the context wants.
+     *
+     * Everything that would otherwise fail deep inside extraction is checked here, so the caller
+     * gets a 400 before the pipeline or Tika is touched: a revision with no locator has nothing to
+     * attach to, and a source key or revision longer than [SourceIdentityBounds] allows would blow
```
jimador, line 306: Add explicit length limits for the source identity and revision before they reach hashing and indexed storage.

jimador: Bounds are answered at the request edge with the broken limit and offending length in the response body, through `SourceIdentityBounds` from #90.

igordayen:
From CODEX:

• Findings:

  - P2 IncrementalPropositionExtraction still projects/wires pre-save propositions
    dice/src/main/kotlin/com/embabel/dice/proposition/extraction/IncrementalPropositionExtraction.kt:467-509 calls result.persist(...),
    then projects and wires the original propsToSave. But PersistablePropositions.persist() uses saveAll(propsToSave) and discards
    every save() return (PersistablePropositions.kt:53-80, PropositionStore.kt:110-112).
    Example: extractor creates proposition fresh-1, but the repository dedups it to existing proposition canonical-1. The REST /extract
    path now uses save()’s returned row, but direct/async extraction still projects and grounds fresh-1, which may not exist in the
    store. This can create projection/grounding relationships for ids later reads cannot resolve. The fix should make this path collect
    canonical saved propositions and pass those to projection/grounding.

  - P3 /extract/file still reports pre-save counts after canonicalization/dedup
    PropositionPipelineController.kt:219-227 saves batch propositions but ignores returned canonical rows, then :259-272 builds chunk
    and total counts from chunkResults, not from what the store kept.
    Example: two chunks extract the same fact; repository dedups both to one stored proposition. The response can still say
    totalPropositions = 2 and each chunk has propositionCount = 1. Since FileExtractResponse only exposes summaries, this is less
    severe than the old /extract id bug, but it is still inconsistent with the “store is authority” behavior now applied to text
    extraction.

  - P3 revision-stable chunk ids are REST-only, but the docs/changelog read broader
    REST calls route chunks through revisionStableChunk() (PropositionPipelineController.kt:346-367). Direct source-aware calls still
    pass sourceId straight through as the chunk id (IncrementalPropositionExtraction.kt:268-315, PropositionPipeline.kt:410-435).
    Example: rememberTextFromSource(..., sourceId = "email:42", revision = "r1") and later revision r2 both ground on "email:42". If
    that is intentional, the changelog/design text should explicitly say revision-stable chunk ids apply to REST only; otherwise
    direct/async entry points need the same derivation.

  No blocking thread-safety issue found in the PR-specific changes. The async event queue/lock path remains serialized, and the new
  event provenance fields are private and only set during construction.


@jimador - please consider rebasing, thanks

igordayen (approved):
@jimador  looks good, thank you

Issue #69, opened 2026-07-24
Return typed product outcomes for batched and linked extraction passes


jimador:
## Goal

Return typed per-product outcomes from bounded extraction plans that support compatible
multi-product calls and linked later passes.

## Context

Facts, decisions, sentiment, and normative claims share proposition lifecycle. Tasks and
recommendations are actionable suggestions with downstream ownership. Batching several
products in one pass and running another pass after entity resolution are independent
capabilities.

## Acceptance criteria

- [ ] `ExtractionRequest` declares source revisions, profile, context, products, speaker
      scope, relationship coverage, pass index, and parent or superseded run.
- [ ] Outcomes are typed per product with stable ids and independent failures.
- [ ] Compatible proposition facets may share one bounded model call.
- [ ] Heterogeneous requests partition before execution.
- [ ] A later pass can consume resolved entities from its parent.
- [ ] Successful siblings remain usable when one product or later pass fails.
- [ ] Budgets, pass limits, ordering, and cycle prevention are explicit.
- [ ] Tests cover meetings, recommendations, partial decode, later-pass replay, and tenants.

## Constraints

- Depends on profile and durable run contracts.
- Preserve serial entity resolution and revision where shared identity requires it.
- DICE returns suggestions but does not own task or notification lifecycle.

## Out of scope

- Downstream effect execution.
- Connector-specific routing.
- An unbounded autonomous extraction agent.

## Original request

Batch compatible meeting and message products where one model call is valuable, preserve
typed failures, and allow a later resolved-entity pass without coupling the two designs.

Issue #67, opened 2026-07-24
Record durable extraction runs and proposition run lineage


jimador:
## Goal

Record durable, tenant-scoped extraction runs and link propositions to every run that
produced or confirmed them without changing source-grounding cardinality.

## Context

Projection and collector runs establish the DICE storage pattern. Extraction needs its
own lifecycle, metadata envelope, indexes, parent/superseded links, and a dedicated
proposition-to-run relation. `ProvenanceEntry` and `DerivedFrom` remain source evidence.

A run may contain zero, one, or many model invocations because of chunking or retries.
Requested configuration, observed execution facts, optional provider-effective facts,
and protected replay evidence must remain distinct.

The design borrows the separation of run/job/input/output from OpenLineage, the
parameters/metrics/context/artifact split from MLflow, requested-versus-response model
facts and opt-in content from OpenTelemetry GenAI, and #64's S3/W3C-PROV-style stable
source plus opaque revision. It uses typed fields rather than generic facet/tag maps.

The resulting evidence supports troubleshooting failures, comparing and evaluating runs,
approximate reproduction, auditing proposition output, and investigating user-context
effects. Experiment/user-context comparisons use tenant-scoped run pages and host-side
grouping; they do not imply new push-down filters or REST endpoints.

## Acceptance criteria

- [ ] `ExtractionRun` records tenant, profile, ordered source revisions, pass, parent/
      superseded lineage, experiment/cohort, prompt/template and schema fingerprints,
      metamodel, extractor/host/runtime, status, timing, counts, and bounded failures.
- [ ] Requested model configuration records portable model/role, temperature, penalties,
      max tokens, top-k/top-p, thinking/selection fingerprints, and timeout without
      serializing provider extension objects.
- [ ] Repeatable invocation records capture only genuinely observed configured service,
      usage, timing, attempt, outcome, and optional provider-response facts.
- [ ] Opaque actor, request, session, personalization, and deployment references support
      audit without retaining PII or arbitrary user-context maps.
- [ ] Raw prompt/source/user-context payloads are absent by default; optional replay
      material is represented only by classified, expiring protected-content references.
- [ ] Replay fidelity is explicit and never promises exact deterministic replay.
- [ ] DICE ships scoped in-memory and Drivine stores, indexes, schema registration,
      lifecycle coordination, invocation correlation, and autoconfiguration.
- [ ] Empty, completed, failed, and cancelled headers persist; retries are idempotent and
      incompatible terminal rewrites are rejected.
[...]

Issue #66, opened 2026-07-24
Carry versioned extraction profiles through compatible DICE entry points


jimador:
## Goal

Carry a trusted, versioned extraction profile through compatible DICE entry points without
changing legacy extraction behavior.

## Context

Perspective describes conversational input, not content policy. DICE needs an opaque
profile reference on `SourceAnalysisContext` and additive `rememberText` inputs so host applications can
select durable policy without connector branches inside DICE.

## Acceptance criteria

- [ ] `ExtractionContentProfileRef` has stable name and version identity.
- [ ] `SourceAnalysisContext` carries an optional profile reference.
- [ ] Additive `rememberText` entry points accept profile and optional run inputs.
- [ ] Legacy callers compile and retain current default behavior.
- [ ] Profiles do not directly select arbitrary providers, models, or credentials.
- [ ] Profile, perspective, schema, and tenant remain independent context dimensions.
- [ ] Java and Kotlin compatibility tests cover absent and present profiles.

## Constraints

- Keep the public interface small and additive.
- DICE carries profile identity; the host authorizes and binds it.
- Preserve current extraction, resolution, and revision ordering.

## Out of scope

- Host profile catalog ownership.
- Durable run persistence.
- Multi-product plans.

## Original request

Give source-independent extraction behavior a stable, versioned identity without encoding
content policy in perspective or hard-coding connectors.

PR #94, opened 2026-08-31
Extraction profiles: versioned identity through compatible entry points

jimador:
PR 5 of the extraction-integrity train (closes the entry-point half of #66; refs #67), stacked on #93. `ExtractionRequest` is the entry-point surface: one value object carrying everything a caller says about an extraction beyond the text and the user — `sourceLocator`, `sourceRevision`, `profile`. Each entry point is exactly two declarations: the signature `main` publishes, byte-identical, plus one overload taking the request. `ExtractionContentProfileRef` names a stable name+version identity for source-independent extraction behavior; opaque by design — DICE carries the identity, the host authorizes and binds what it names, and profiles select no provider, model, or credential. The request validates its own coupling in `init`: a revision requires a locator whose key it matches, so a mismatched pair can never reach an entry point.

- **The request object replaces the widening overloads.** All six in-train wide overloads are deleted; `rememberTextFromSource` and `rememberFileFromSource` no longer exist as names, so the wide forms never ship in a release and #69 has nothing to deprecate. The next dimension extraction learns about arrives as a field on the request while the entry-point signatures stay put.
- **Profile is carried once.** The request is the single entry-point door onto `SourceAnalysisContext`; the context keeps `profile` as the downstream read point — a host `PropositionExtractor` receives the whole context on every `extract(chunk, context)` call, and the run record reads it there once the coordinator lands.
- **No no-op parameters.** The `currentRun` parameter that was accepted and then dropped at persistence is removed here; a reflection sweep fails the build if any request member mentions a run before the consuming write exists. It returns as the request's fourth field on #101, together with the lineage write that consumes it.

**Breaking changes: none — including for subclasses.** Every pre-request entry-point signature stays present, open, and reached, pinned four ways: the declared descriptor set, non-finality assertions, a subclass override called on every published arity (positional and named), and Java interop with `@Override` compile proofs. New constructor parameters follow the stated Kotlin-synthetics boundary.

**Opt-in and status:** **EXPERIMENTAL** (`@ApiStatus.Experimental`). The trigger is passing an `ExtractionRequest`; legacy calls behave byte-identically, and `ExtractionRequest.NONE` dispatches exactly as the legacy call it resembles, so a subclass overriding only the six-argument `rememberText` still intercepts file ingestion.

**The dispatch shape** — every call funnels to one terminal method, and legacy overrides stay reached:

```mermaid
flowchart TD
    A[legacy rememberText] --> W[rememberText with request]
    B[legacy rememberFile] --> X[rememberFile with request]
    X --> W
    E[async event paths] --> W
    W --> C[buildContext]
    C --> D[SourceAnalysisContext: locator, revision, profile, perspective, schema, tenant]
    S[a subclass override of a legacy signature stays reached] -.-> X
```

Overridable and reached are different claims, so they are tested separately.


dice/src/main/kotlin/com/embabel/dice/proposition/extraction/IncrementalPropositionExtraction.kt
```diff
@@ -461,6 +625,14 @@ open class IncrementalPropositionExtraction @JvmOverloads constructor(
         if (sourceRevision != null) {
             ctx = ctx.withSourceRevision(sourceRevision)
         }
+        // Carried, never consulted. Nothing downstream of here reads either one — that is what
+        // "DICE holds profile identity and the host binds policy" means in code.
+        if (profile != null) {
+            ctx = ctx.withProfile(profile)
+        }
+        if (currentRun != null) {
+            ctx = ctx.withCurrentRun(currentRun)
```
jimador: This accepts run attribution and then drops it when `persistAndProject` saves the propositions. Can we add `currentRun` with the proposition lineage write that consumes it so callers do not depend on a no-op parameter?

jimador: Resolved in #101: a run present now switches persistence, projection and grounding onto the canonical propositions the repository returns, and the lineage write consumes it when a PropositionRunLinkStore is bound. The stale 'carried, never consulted' docs at this site and its three siblings are rewritten there too.

jimador: Removed the parameter, and `ExtractionRunRef` with it. Run attribution arrives on the lineage PR alongside the proposition lineage write that consumes it, so no caller here depends on a no-op.

jimador:
Two design points on the entry-point side.

`profile` is carried and never read. DICE consults it nowhere, no host catalog exists, and REST has no field for it. As shipped it is an attribution tag, and the run record already carries `profile`. Carrying it separately on `SourceAnalysisContext` and on every `remember*` overload duplicates it.

The `remember*` surface is growing one parameter per feature. After this stack `rememberTextFromSource` has an eight-argument and a ten-argument form, all four entry points carry a wide overload, and `SourceAnalysisContext` has three new optional fields. Slice 11 in the train design introduces `ExtractionRequest` with source revisions, profile, context, products, pass index and parent run. That is the parameter object these overloads approximate. Introduce the request object now, put revision, profile and run on it, and keep every legacy signature exactly as it is. The wide overloads then never ship in a release and #69 has nothing to deprecate.

jimador:
The request object is in now. Landed as "Introduce ExtractionRequest as the entry-point surface".

**The duplication is gone.** `ExtractionRequest` is the only door from the entry points onto the context, and `profile` rides it once. `SourceAnalysisContext.profile` stays as the single downstream read point — a host-supplied `PropositionExtractor` receives the whole context on every `extract(chunk, context)` call, and the run record reads it there when the coordinator lands. The run ref that had no consumer had already been cut ("Carry profile identity without promising run attribution"); it returned as the request's fourth field on #101, together with the lineage write that consumes it.

**The `remember*` surface stops growing.** Each entry point is exactly two declarations: the signature `main` publishes, byte-identical, plus one overload taking the request. All six in-train wide overloads are deleted — `rememberTextFromSource` and `rememberFileFromSource` no longer exist as names — so the wide forms never ship in a release and #69 has nothing to deprecate. The request carries locator, revision and profile, and validates the revision/locator pairing in its own `init`, so a mismatched pair can never reach an entry point. Legacy signatures are pinned four ways: declared descriptor set, non-finality, a subclass override reached on every published arity, and Java interop with `@Override` compile proofs.

The next dimension extraction learns about arrives as a field on the request while the entry-point signatures stay where they are.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRequest.kt
```diff
@@ -0,0 +1,107 @@
...
+ *   source gets richer grounding than the content-hash fallback.
+ * @property sourceRevision the provider's own identifier for the version of [sourceLocator] this
+ *   run reads. Supplying one asserts that the revision covers the whole text or the whole file
+ *   being extracted; DICE reads material as one aggregate and cannot work that out for itself.
+ * @property profile the host's content-profile identity for this extraction. DICE carries it and
+ *   does nothing else with it: no provider, model, or credential is chosen from it, and extraction
+ *   runs exactly as it would without one.
+ */
+@ApiStatus.Experimental
+data class ExtractionRequest @JvmOverloads constructor(
```
jimador, line 54: Two design points on the entry-point side.

`profile` was carried and never read. DICE consults it nowhere, no host catalog exists, and REST has no field for it. As shipped it is an attribution tag, and the run record already carries `profile`. Carrying it separately on `SourceAnalysisContext` and on every `remember*` overload duplicates it.

The `remember*` surface was growing one parameter per feature. After this stack `rememberTextFromSource` had an eight-argument and a ten-argument form, all four entry points carried a wide overload, and `SourceAnalysisContext` had three new optional fields. Slice 11 in the train design introduces `ExtractionRequest` with source revisions, profile, context, products, pass index and parent run. That is the parameter object these overloads approximate. Introduce the request object now, put revision, profile and run on it, and keep every legacy signature exactly as it is. The wide overloads then never ship in a release and #69 has nothing to deprecate.

jimador: The request object is in. Landed as "Introduce ExtractionRequest as the entry-point surface".

**The duplication is gone.** `ExtractionRequest` is the only door from the entry points onto the context, and `profile` rides it once. `SourceAnalysisContext.profile` stays as the single downstream read point: a host-supplied `PropositionExtractor` receives the whole context on every `extract(chunk, context)` call, and the run record reads it there when the coordinator lands. The run ref that had no consumer had already been cut; it returned as the request's fourth field on #101, together with the lineage write that consumes it.

**The `remember*` surface stops growing.** Each entry point is exactly two declarations: the signature `main` publishes, byte-identical, plus one overload taking the request. All six in-train wide overloads are deleted, so the wide forms never ship in a release and #69 has nothing to deprecate. The request carries locator, revision and profile, and validates the revision/locator pairing in its own `init`. Legacy signatures are pinned four ways: declared descriptor set, non-finality, a subclass override reached on every published arity, and Java interop with `@Override` compile proofs.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRequest.kt
```diff
@@ -0,0 +1,107 @@
...
+    val sourceLocator: SourceLocator? = null,
+    val sourceRevision: SourceRevisionRef? = null,
+    val profile: ExtractionContentProfileRef? = null,
+) {
+
+    init {
+        sourceRevision?.let { revision ->
```
igordayen, line 61: is data class validation on ```init ```standard practice?

Suppose, theoretically, a secondary constructor were added; would ```init``` be executed?

If the pattern gets used in many places, perhaps introduce ValidationAware interface?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/IncrementalPropositionExtraction.kt
```diff
@@ -455,12 +448,13 @@ open class IncrementalPropositionExtraction @JvmOverloads constructor(
...
         }
-        if (sourceLocator != null) {
-            ctx = ctx.withSourceLocator(sourceLocator)
-        }
-        if (sourceRevision != null) {
-            ctx = ctx.withSourceRevision(sourceRevision)
-        }
+        // The request is the one door everything the caller asked for comes through, so the
+        // context and the request always agree about a call.
+        request.sourceLocator?.let { ctx = ctx.withSourceLocator(it) }
```
igordayen, line 453: if im not mistken ```sourceLocator``` has been already validated in ```init```

igordayen (commented):
@jimador - looks good, minor inquiries on data classes validation / usage.  thanks

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRequest.kt
jimador: `init` runs for every construction path. A secondary constructor in Kotlin has to delegate to the primary one, and the primary's `init` blocks run as part of that delegation, so there is no constructor that skips it. `copy()` goes through the primary constructor too. That is why the check lives there and not in a factory: there is no way to get an instance that was not checked.

On a `ValidationAware` interface: what would it add? The check has to run at construction to keep an invalid instance from existing at all; an interface method a caller has to remember to invoke is weaker than that. Where the same rule shows up in several types (the bounds in `SourceIdentityBounds` are the case today) the shared piece is the rule, and each type's `init` calls it. Happy to revisit if a use case needs validation after construction.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/IncrementalPropositionExtraction.kt
jimador: Different job. `init` checks that a revision, when present, names the same source the locator does. The `?.let` here is null handling: `sourceLocator` is optional on the request, and the context only gets one when the caller supplied one. Nothing is validated twice.

igordayen (approved):
@jimador - looks good, subject to rebasing, thanks

PR #95, opened 2026-08-31
Extraction runs: the durable run model

jimador:
PR 6 of the extraction-integrity train (refs #67 — the model half), stacked on #94. The durable extraction-run record, as pure JVM value types. `ExtractionRun` is keyed by (`ContextId`, `ExtractionRunRef`) and carries profile, ordered source revisions, pass, parent/superseded refs plus a denormalized root-run ref (OpenLineage's ParentRunFacet pattern — a whole-lineage lookup becomes one indexed read), experiment/cohort refs, prompt/schema/metamodel fingerprints, runtime identity, status, timing, counts, and bounded failures. Requested and observed model facts are separate types with disjoint fields, so a guessed provider-effective value has nowhere to live. Invocation identity is the ordinal in a call plan fixed before any call fires; attempts number retries; completion order never mints. Opaque pseudonymous refs carry actor/request/session identity with serialized-row privacy tests.

- **Failures speak a closed vocabulary.** `ExtractionFailure` has no `String` parameter, property or field anywhere in its reachable shape and no factory taking a `Throwable`: `code` (11 values), `stage`, a validated `providerStatus`, one typed `measure` with its unit, `at`, and the invocation it names. `detail`, `fromThrowable` and the sanitize/cause-chain helpers are removed, so prompt content, PII, credentials and protected text have nothing shaped to hold them and cannot reach durable storage through this type.
- **Run recording aligns with the source-revision contract.** The revision bound this type imposed is gone — a run accepts whatever `SourceRevisionRef` accepts, so extraction and revision queries can no longer succeed where run recording then fails.
- **The lineage root is derived.** The constructor is private and `childOf` derives the root from the parent's own lineage, so a root that contradicts its parent has no public path in; the KDoc records the reflection residual honestly.
- `metamodelFingerprint` keeps its field and its KDoc names its one writer: the extraction coordinator, resolving the host's `DeclaredSchemaSource` — per the decision that per-proposition schema attribution answers through run lineage.

**Breaking changes: none.** New types only. The failure-vocabulary rework reshapes types introduced earlier in this same unreleased train; no released surface moves.

**Opt-in and status:** **EXPERIMENTAL** — every type carries `@ApiStatus.Experimental`. Nothing here executes; the store contract, the Drivine store, and lineage stamping follow as #98, #99 and #101.

```mermaid
erDiagram
    ExtractionRun ||--o{ ExtractionInvocationRecord : "invocations, identity fixed before any call"
    ExtractionRun ||--o{ ExtractionFailure : "bounded, typed vocabulary, no free text"
    ExtractionRun ||--|| ExtractionRunLineage : "runRef, root, parent, pass"
    ExtractionRun }o--|| ExtractionContentProfileRef : "profile"
    ExtractionRun }o--o{ SourceRevisionRef : "ordered source revisions"
    ExtractionFailure }o--o| ExtractionInvocationRecord : "may name only an attempt the run records"
```

The denormalized root is what makes a whole-lineage lookup one indexed read.

### DIFF
--- a/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
@@ -1,0 +1,177 @@
+/*
+ * Copyright 2024-2026 Embabel Pty Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.proposition.extraction
+
+import org.jetbrains.annotations.ApiStatus
+import java.time.Instant
+
+/**
+ * What kind of thing went wrong, from a fixed list.
+ *
+ * The code is what a query groups by and what an alert fires on, so it is an enum rather than a
+ * string a caller invents per site. Anything that does not fit is [INTERNAL] or [UNCLASSIFIED],
+ * and a code that keeps getting used for the wrong thing is a signal to add one.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ */
+@ApiStatus.Experimental
+enum class ExtractionFailureCode {
+
+    /** The model or its service could not be reached, or refused to serve the request. */
+    MODEL_UNAVAILABLE,
+
+    /** The call was still outstanding when the configured timeout expired. */
+    MODEL_TIMEOUT,
+
+    /** The provider declined to answer — a safety refusal, a content filter, a policy block. */
+    MODEL_REFUSED,
+
+    /** The provider rejected the call for quota or rate reasons. */
+    RATE_LIMITED,
+
+    /** A response arrived and could not be parsed into the expected shape. */
+    DECODE_FAILED,
+
+    /** A response parsed and then broke the schema or the metamodel it had to satisfy. */
+    SCHEMA_VIOLATION,
+
+    /** The source material could not be read at the revision the run was asked for. */
+    SOURCE_UNAVAILABLE,
+
+    /** Extraction produced results and storing them failed. */
+    PERSISTENCE_FAILED,
+
+    /** The run, or this part of it, was stopped before it finished. */
+    CANCELLED,
+
+    /** A defect on DICE's side of the boundary. */
+    INTERNAL,
+
+    /** Nothing above fits, and the caller would rather record the failure than force a code. */
+    UNCLASSIFIED,
+}
+
+/**
+ * One thing that went wrong during a run: a code, a short detail, when, and which invocation.
+ *
+ * A failure record is evidence, and evidence about a failure is where source text leaks. A model
+ * provider routinely quotes the prompt back in its exception message, a decode error carries the
+ * fragment it choked on, and both end up in a stored run header if someone writes
+ * `e.message` into one. So the DICE-minted path, [fromThrowable], never reads
+ * `Throwable.message` at all: it records the exception class names down the cause chain and
+ * nothing else, which cannot contain source text because it never touched any.
+ *
+ * [of] exists for the case where a caller genuinely knows something useful ("chunk 3 of 12
+ * exceeded the token budget"). DICE cannot check what a caller puts there. It bounds it, flattens
+ * it to a single line so a pasted stack trace does not fit, and the contract is that the caller
+ * supplies a classification rather than a payload.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property code How the failure is classified
+ * @property detail Short, single-line, sanitized explanation. Empty when the code says it all.
+ * @property at When the failure was recorded
+ * @property invocation The invocation and attempt this failure belongs to, or null for a failure
+ *   that happened outside any model call
+ */
+@ApiStatus.Experimental
+data class ExtractionFailure @JvmOverloads constructor(
+    val code: ExtractionFailureCode,
+    val detail: String = "",
+    val at: Instant = Instant.now(),
+    val invocation: ExtractionInvocationId? = null,
+) {
+
+    init {
+        require(detail.length <= ExtractionRunLimits.MAX_FAILURE_DETAIL_LENGTH) {
+            "detail must be at most ${ExtractionRunLimits.MAX_FAILURE_DETAIL_LENGTH} characters, " +
+                "was ${detail.length}"
+        }
+        require(detail.none { it == '\n' || it == '\r' }) {
+            "detail must be a single line; a multi-line detail is a stack trace or a quoted payload"
+        }
+    }
+
+    companion object {
+
+        /** How many links of a cause chain [fromThrowable] walks. */
+        const val MAX_CAUSE_CHAIN: Int = 5
+
+        /**
+         * Records a failure with a detail the caller wrote.
+         *
+         * Whitespace collapses to single spaces and the result is clipped to
+         * [ExtractionRunLimits.MAX_FAILURE_DETAIL_LENGTH]. The caller is responsible for the
+         * detail holding no source text, no prompt, and no personal data; nothing here can verify
+         * that. [fromThrowable] is the path to use when the detail would have come from an
+         * exception.
+         */
+        @JvmStatic
+        @JvmOverloads
+        fun of(
+            code: ExtractionFailureCode,
+            detail: String,
+            at: Instant = Instant.now(),
+            invocation: ExtractionInvocationId? = null,
+        ): ExtractionFailure = ExtractionFailure(
+            code = code,
+            detail = sanitize(detail),
+            at = at,
+            invocation = invocation,
+        )
+
+        /**
+         * Records a failure from a throwable, using its class names and nothing else.
+         *
+         * The detail is the exception class name, then up to [MAX_CAUSE_CHAIN] causes joined by
+         * ` <- `. `Throwable.message`, suppressed exceptions, and the stack trace are all
+         * untouched, so the run header cannot pick up the prompt, the response body, or the
+         * fragment that failed to parse.
+         *
+         * A host that wants the message keeps it in its own logs, where retention and access are
+         * its to set.
+         */
+        @JvmStatic
+        @JvmOverloads
+        fun fromThrowable(
+            code: ExtractionFailureCode,
+            throwable: Throwable,
+            at: Instant = Instant.now(),
+            invocation: ExtractionInvocationId? = null,
+        ): ExtractionFailure = ExtractionFailure(
+            code = code,
+            detail = sanitize(causeChain(throwable).joinToString(" <- ")),
+            at = at,
+            invocation = invocation,
+        )
+
+        private fun causeChain(throwable: Throwable): List<String> {
+            val names = mutableListOf<String>()
+            var current: Throwable? = throwable
+            val seen = mutableSetOf<Throwable>()
+            while (current != null && names.size < MAX_CAUSE_CHAIN && seen.add(current)) {
+                names += current.javaClass.name
+                current = current.cause
+            }
+            return names
+        }
+
+        private fun sanitize(detail: String): String =
+            detail.replace(WHITESPACE, " ").trim().take(ExtractionRunLimits.MAX_FAILURE_DETAIL_LENGTH)
+
+        private val WHITESPACE = Regex("\\s+")
+    }
+}
--- a/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
@@ -21,9 +21,9 @@ import java.time.Instant
 /**
  * What kind of thing went wrong, from a fixed list.
  *
- * The code is what a query groups by and what an alert fires on, so it is an enum rather than a
- * string a caller invents per site. Anything that does not fit is [INTERNAL] or [UNCLASSIFIED],
- * and a code that keeps getting used for the wrong thing is a signal to add one.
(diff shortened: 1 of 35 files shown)


dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRun.kt
```diff
@@ -0,0 +1,327 @@
...
+            "sourceRevisions must be distinct; a run reads each source revision once"
+        }
+        // SourceRevisionRef predates the run model's cap rule and validates non-blank only, so the
+        // bound is applied where the run stores it. Moving it onto the type is a follow-up.
+        this.sourceRevisions.forEach { revision ->
+            require(revision.sourceKey.length <= ExtractionRunLimits.MAX_SOURCE_KEY_LENGTH) {
+                "sourceKey must be at most ${ExtractionRunLimits.MAX_SOURCE_KEY_LENGTH} characters, " +
+                    "was ${revision.sourceKey.length}"
+            }
+            require(revision.sourceRevision.length <= ExtractionRunLimits.MAX_IDENTIFIER_LENGTH) {
```
jimador: `SourceRevisionRef` accepts revisions longer than this, so extraction and revision queries can succeed and run recording then fails. Can we keep the bound aligned with the source revision contract until a real store constraint requires one?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunLineage.kt
```diff
@@ -0,0 +1,120 @@
...
+ * @property rootRunRef The oldest run in this lineage, which is [runRef] itself when there is no
+ *   parent
+ * @property parentRunRef The run this one continues from, or null
+ * @property supersedesRunRef The run this one replaces, or null
+ * @property passIndex Which pass over the material this is, counting from zero
+ */
+@ApiStatus.Experimental
+data class ExtractionRunLineage @JvmOverloads constructor(
+    val runRef: ExtractionRunRef,
+    val rootRunRef: ExtractionRunRef,
```
jimador, line 71: Any non-self root passes this constructor, and the store never verifies it against the parent. Can we derive the root from the stored parent or defer this denormalization until a query needs it?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRun.kt
jimador: Dropped the cap. `sourceKey` and `sourceRevision` carry no length check here or on `SourceRevisionRef`: a revision is opaque and non-blank per `docs/design/source-revisions.md`, and a run accepts whatever that contract accepts. A bound can return when a real store constraint requires one.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunLineage.kt
jimador: The constructor is private and `childOf` derives the root from the parent's own lineage, so a root disagreeing with the parent chain has no public path in. The KDoc records the residual: Kotlin reflection and a Jackson deserializer still reach the private constructor, and nothing serializes this type today.

jimador:
`ExtractionFailure` is not sanitized by construction. Its public constructor and `of` factory accept arbitrary source text, and the Drivine mapper persists that text verbatim. Truncating it and flattening line breaks does not prevent prompt content, PII, credentials, or protected text from entering durable storage.

Make the persisted diagnostic a controlled structured vocabulary, with any detailed material represented by a protected reference. Add canary tests proving raw source text and secrets cannot cross this boundary.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
```diff
@@ -0,0 +1,251 @@
...
+ * What kind of thing went wrong, from a fixed list.
+ *
+ * The code is what a query groups by and what an alert fires on, so it comes from a fixed list
+ * that no caller can extend at a call site. Anything that does not fit is [INTERNAL] or
+ * [UNCLASSIFIED], and a code that keeps getting used for the wrong thing is a signal to add one.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ */
+@ApiStatus.Experimental
+enum class ExtractionFailureCode {
```
jimador, line 31: `ExtractionFailure` was not sanitized by construction. Its public constructor and `of` factory accepted arbitrary source text, and the Drivine mapper persisted that text verbatim. Truncating it and flattening line breaks does not prevent prompt content, PII, credentials, or protected text from entering durable storage.

Make the persisted diagnostic a controlled structured vocabulary, with any detailed material represented by a protected reference. Add canary tests proving raw source text and secrets cannot cross this boundary.

jimador: Failures speak a closed vocabulary. `ExtractionFailure` has no `String` parameter, property or field anywhere in its reachable shape and no factory taking a `Throwable`: `code` (11 values), `stage`, a validated `providerStatus`, one typed `measure` with its unit, `at`, and the invocation it names. `detail`, `fromThrowable` and the sanitize/cause-chain helpers are removed, so prompt content, PII, credentials and protected text have nothing shaped to hold them and cannot reach durable storage through this type. The Drivine mapper persists the vocabulary (#99, "Persist the failure vocabulary and announce transitions from Drivine").

docs/design/extraction-runs.md
```diff
@@ -0,0 +1,404 @@
...
+    EXP --> OP
+```
+
+`ExtractionRunKey` is the tenant-qualified identity: a run id is host-minted and DICE never
+assumes it is unique across tenants, so the `ContextId` travels with it. Two tenants that both
+mint `run-1` have two runs.
+
+## How a run joins to the metamodel
+
+One field carries the join. `ExtractionRunFingerprints.metamodelFingerprint` holds the declared
```
jimador, line 50: The join with the metamodel stack was undefined. `ExtractionRun.fingerprints.metamodelFingerprint` was a pass-through string. On the other stack `dice.metamodel.version` was a proposition metadata key nothing writes. Both describe the same fact, which schema declaration a claim was extracted under. Decide once. The run carries the `DeclaredSchema` stamp hash, the coordinator resolves it from the `DeclaredSchemaSource` bean when one is present, and per-proposition schema attribution is answered through `PRODUCED_BY_RUN`. The metadata key then goes, or becomes a denormalised copy the coordinator writes.

Then: the decision was in KDoc only. `ExtractionRunEnvelope.kt` said the coordinator writes `metamodelFingerprint` from the host's `DeclaredSchemaSource`, and this doc never mentioned the field. Put the decision here, in the same words as `metamodel-versioning.md` on the other stack, so the two docs describe one mechanism.

jimador: **One attribution mechanism: the run.** On #83 the metadata key is removed ("Answer schema attribution through run lineage"); the run carries the declared schema's `contentHash` and per-proposition attribution answers through `PRODUCED_BY_RUN`. `metamodelFingerprint` keeps its field, and its KDoc names its one writer, the coordinator resolving the host's `DeclaredSchemaSource`, so nothing else may fill it.

7d8a8dd (on #95's branch) adds "How a run joins to the metamodel" to this doc, in the same words as `metamodel-versioning.md`: the coordinator resolves the host's `DeclaredSchemaSource` and stamps `metamodelFingerprint` on the run record.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
```diff
@@ -0,0 +1,251 @@
...
+
+    /** The model or its service could not be reached, or refused to serve the request. */
+    MODEL_UNAVAILABLE,
```
igordayen, line 34: MODEL_UNAVAILABLE("description")?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
```diff
@@ -0,0 +1,251 @@
...
+ * @property value The number
+ */
+@ApiStatus.Experimental
+data class ExtractionFailureMeasure(
+    val quantity: ExtractionFailureQuantity,
+    val value: Long,
+) {
+
+    init {
+        require(value >= 0) { "$quantity must not be negative, was $value" }
```
igordayen, line 167: same validation pattern comments,  on PR#94

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionInvocationRecord.kt
```diff
@@ -0,0 +1,298 @@
...
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property invocationIndex Ordinal of this call in the run's plan, counting from zero
+ * @property attempt Which try at that call this is, counting from one
+ */
+@ApiStatus.Experimental
+data class ExtractionInvocationId(
+    val invocationIndex: Int,
+    val attempt: Int,
```
igordayen, line 43: is there any external config for max attempts

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionInvocationRecord.kt
```diff
@@ -0,0 +1,298 @@
...
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property inputTokens Tokens the provider counted on the way in
+ * @property outputTokens Tokens the provider counted on the way out
+ * @property totalTokens The provider's own total, when it gave one
+ * @property cachedInputTokens Input tokens the provider served from its cache
+ * @property reasoningTokens Output tokens the provider attributed to reasoning
+ */
+@ApiStatus.Experimental
+data class ExtractionModelUsage @JvmOverloads constructor(
```
igordayen, line 113: Maybe extend Usage (from core)?, Seems conceptually similar

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRequestedModelConfig.kt
```diff
@@ -0,0 +1,118 @@
...
+ * @property maxTokens Requested output-token ceiling
+ * @property presencePenalty Requested presence penalty
+ * @property frequencyPenalty Requested frequency penalty
+ * @property thinkingFingerprint Opaque digest of the reasoning or thinking configuration
+ * @property selectionFingerprint Opaque digest of how the host chose this model and these settings
+ * @property timeout How long the caller was prepared to wait
+ */
+@ApiStatus.Experimental
+data class ExtractionRequestedModelConfig @JvmOverloads constructor(
+    val modelRole: String? = null,
```
igordayen, line 65: please consider embedding LlmHyperparameters

docs/design/extraction-runs.md
```diff
@@ -0,0 +1,404 @@
+# Extraction runs: what produced a claim, recorded without holding what it read
+
+An extraction run is the durable record of one execution: which profile, prompt, schema and
```
igordayen, line 3: conceptually similar to what-would-be-episode audit

igordayen (commented):
@jimador - looks vey good, few comments to consider, thanks

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
jimador: Kept the enum bare on purpose. The constant's name is the wire value: it is what a query groups by, what the REST surface returns, and what an alert matches on, so it has to be stable and language-neutral. The prose lives in KDoc next to each constant, which is where a reader looks it up and where it can change without touching stored data. A description property would be a second copy of that prose that callers would start rendering, and then we could not reword it freely.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionFailure.kt
jimador: Same answer as on #94: `init` runs for every construction path, including secondary constructors and `copy()`, so the invariant holds for every instance. The rule itself is one line per type; there is nothing yet worth pulling into a shared interface.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionInvocationRecord.kt
jimador: No, and on purpose in this slice. The record is descriptive: `attempt` says which try this was, and the store keeps one row per `(runId, index, attempt)`. Who retries, and how many times, is the caller's policy (the pipeline's retry settings, or the framework's), and the run store neither enforces a ceiling nor knows one. If a ceiling ever needs to be recorded, it belongs on the run header as an observed fact of the plan, not as configuration of the store. Noted as a follow-up for the coordinator slice.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionInvocationRecord.kt
jimador: Looked at extending it. Core `Usage` is a final data class and it carries `nativeUsage: Any?`, the SDK object this record exists to not store; it also has no cached-input or reasoning counts. So the record stays its own type and gains `ExtractionModelUsage.from(Usage)` in bb6fdfa0, which maps prompt, completion and total tokens and leaves the two extra counts null.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRequestedModelConfig.kt
jimador: Done in 1c7d081e. `ExtractionRequestedModelConfig` now implements `LlmHyperparameters`; the six fields already matched the interface member for member, so it is the same stored shape read back as the framework's type, plus `from(LlmOptions)` to build one from what a host handed the model. `thinking` stays a fingerprint because `Thinking` is not a stored value.

docs/design/extraction-runs.md
jimador: Agreed, and deliberately so. An episode audit is the consumer-side view of the same thing: what ran, with what, and what came out. The run model is the producer-side record DICE can write without knowing anything about the host's episode concept, and `profile` plus the run lineage are the join points a host would use to fold runs into its own audit. I added a short note in the design doc naming that relationship so the next reader does not have to rediscover it.

igordayen (approved):
@jimador - looks good, subject to rebase. thanks

PR #98, opened 2026-08-31
Extraction runs: lifecycle machine and store contract

jimador:
PR 7 of the extraction-integrity train (refs #67), stacked on #95. The run lifecycle and store contract, in `dice` core. The state machine lives in one place: `save` accepts only RUNNING runs, and `ExtractionRunTransition.applyTo` under store compare-and-set is the only way a run reaches COMPLETED/FAILED/CANCELLED — nothing can manufacture a terminal run without a store. Idempotency has teeth: replaying the same terminal payload succeeds; an incompatible rewrite is rejected by insert-or-compare on a SHA-256 fingerprint of the transition payload, deterministic and golden-literal-locked. `ExtractionRunStore` gives tenant-scoped bounded pages (scope before limit), bounded cycle-safe chain walks, whole-lineage lookup in one read off the denormalized root, and invocation persistence that survives header saves. The cross-backend contract suite carries the discriminating cases — interleaved replay, lineage disagreement, the terminal-by-terminal matrix, a transition race, invocation preservation — so the Drivine store inherits the semantics.

- **The terminal fingerprint hashes transition identity only** — `(status, finishedAt)`. Counts and failures travel as data on the transition and still participate in its equality, so #69's typed product outcomes change no persisted fingerprint format. `TERMINAL_VERSION` moved to `xrun-terminal:v2` once, before any release, and the design doc states the choice.
- **Applied transitions are announced.** `ExtractionRunTransitioned` is a new `DiceEvent` carrying the terminal run; the in-memory store takes a `DiceEventListener` collaborator and emits after the write lands, gated on the result being applied. Nothing is wired automatically — the audit projection and host listeners have their subscription point.
- **Header writes carry a version under compare-and-set.** A stale save used to rebuild every header field, so it could erase revisions, counts, failures, profile or fingerprints another writer had stored; it is now rejected. Terminal invocation records are locked the same way: an identical replay succeeds, an incompatible rewrite is refused on the fingerprint.
- `ProtectedContentRef` is deferred out: DICE had no writer, reader, or retention behavior for it; it returns with the first runtime path that uses it. The doc states the reference as it now stands.

**Breaking changes: none.** New types only; the header-version parameter's binary consequences are enumerated in the CHANGELOG (which descriptors it preserves, which it moves, and what happens to a caller that never passes one).

**Opt-in and status:** **EXPERIMENTAL**, `@ApiStatus.Experimental` on every type. Nothing calls the store yet; the coordinator is the first real caller and is sequenced next.

```mermaid
stateDiagram-v2
    [*] --> RUNNING: save accepts RUNNING only
    RUNNING --> RUNNING: save under versioned compare-and-set
    RUNNING --> SUCCEEDED: transition
    RUNNING --> FAILED: transition
    RUNNING --> CANCELLED: transition
    SUCCEEDED --> SUCCEEDED: identical replay succeeds
    FAILED --> FAILED: identical replay succeeds
    CANCELLED --> CANCELLED: identical replay succeeds
```

**Two write doors, and what each promises:**

```mermaid
flowchart TD
    A[save: whole header] --> V{version matches stored?}
    V -- no --> R[rejected, stored row untouched]
    V -- yes --> W[header replaced, invocations merged by identity]
    B[recordInvocation: one child row] --> T{record already terminal?}
    T -- yes --> F{payload fingerprint identical?}
    F -- yes --> K[accepted as replay, nothing moves]
    F -- no --> R2[rejected]
    T -- no --> M[child row inserted or compared]
    W --> E[applied terminal transition emits ExtractionRunTransitioned]
```

Closed value sets are exercised by iterating their own `entries`, so a newly declared constant is covered the day it is added.

- **`ExtractionRunStore.save` states the returned version for each kind of save.** A first save returns version 0 (the version the caller named, now confirmed by the store), an accepted update the stored version plus one, and a replay the stored version unchanged. The KDoc used to say a first save returned the caller's version plus one. The contract test pins all three.

### DIFF
--- a/dice/src/main/kotlin/com/embabel/dice/common/DiceEvent.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/common/DiceEvent.kt
@@ -20,7 +20,9 @@ import com.embabel.common.core.types.Timestamped
 import com.embabel.dice.pipeline.PropositionExtractionStats
 import com.embabel.dice.proposition.Proposition
 import com.embabel.dice.proposition.PropositionStatus
+import com.embabel.dice.proposition.extraction.ExtractionRun
 import com.fasterxml.jackson.annotation.JsonTypeInfo
+import org.jetbrains.annotations.ApiStatus
 import java.time.Instant
 
 /**
@@ -157,6 +159,30 @@ data class ProjectionBatchCompleted @JvmOverloads constructor(
  */
 data class ExtractionBatchCompleted @JvmOverloads constructor(
     val stats: PropositionExtractionStats,
+    override val timestamp: Instant = Instant.now(),
+) : DiceEvent
+
+/**
+ * An extraction run ended, and this is the call that ended it.
+ *
+ * It fires once per run, from the store, after the terminal write has landed. A coordinator
+ * retrying a terminal write it never saw the answer to gets a replay, and a replay fires nothing —
+ * so a listener counting finished runs, or kicking off work behind one, sees each run once however
+ * many times its terminal write was sent.
+ *
+ * The run carries everything there is to know about how it ended: the tenant, the lineage, the
+ * terminal status, the finish time, the counts, and the failures in their closed vocabulary. There
+ * is no source text or provider message anywhere in it, so this event is safe to hand to a listener
+ * that logs or forwards what it receives.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property run The run in its terminal state.
+ * @property timestamp When the event was created.
+ */
+@ApiStatus.Experimental
+data class ExtractionRunTransitioned @JvmOverloads constructor(
+    val run: ExtractionRun,
     override val timestamp: Instant = Instant.now(),
 ) : DiceEvent
 
--- a/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunTransition.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunTransition.kt
@@ -37,8 +37,12 @@ import java.util.Collections
  * for the same reason from the other side: a run that retried past a failed attempt and finished
  * still happened.
  *
- * [fingerprint] is what a store compares a retry against; see [ExtractionRunFingerprint] for the
- * encoding and why it is specified rather than left to a serializer.
+ * **[fingerprint] names the write, and the counts and failures ride beside it.** A store compares a
+ * repeated terminal write against the digest of [status] and [finishedAt] alone. Two transitions
+ * that agree on those are the same write, so the second replays and the run keeps the counts and
+ * failures the first one delivered — a run's outcome is written once. See
+ * [ExtractionRunFingerprint] for the encoding, and for why a payload that grows leaves the digest
+ * where it is.
  *
  * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
  *
@@ -74,9 +78,11 @@ class ExtractionRunTransition @JvmOverloads constructor(
 
     /**
      * The digest a store compares a repeated terminal write against, computed once at construction.
+     *
+     * It covers this transition's identity — [status] and [finishedAt] — and nothing else. See
+     * [ExtractionRunFingerprint] for why [counts] and [failures] travel as data beside it.
      */
-    val fingerprint: String =
-        ExtractionRunFingerprint.ofTerminal(status, finishedAt, counts, this.failures)
+    val fingerprint: String = ExtractionRunFingerprint.ofTerminal(status, finishedAt)
 
     /**
      * Derives the terminal run this transition produces from the running one it is applied to.
--- a/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunTransition.kt
+++ b/dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunTransition.kt
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2024-2026 Embabel Pty Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.proposition.extraction
+
+import org.jetbrains.annotations.ApiStatus
+import java.time.Instant
+import java.util.Collections
+
+/**
+ * One terminal write against a run: the state it ends in, when it ended, and the counts and
+ * failures that go with it.
+ *
+ * This type is how a run becomes a different run. [ExtractionRun] publishes no `withStatus`, no
+ * `finished()` and no `copy`, so [applyTo] is the only place a terminal run is derived. A store
+ * applies it under compare-and-set; nothing else may.
+ *
+ * **Null means keep, a value means replace.** [counts] and [failures] are both nullable and both
+ * follow the same rule, so a coordinator that only knows the run stopped can say that and nothing
+ * more. An empty [failures] list is a value: it replaces whatever the run accumulated with none.
+ *
+ * **A `FAILED` transition need not carry a failure.** A run can stop on something the coordinator
+ * classifies at the run level with no per-attempt detail, and requiring the pairing would make an
+ * honest "we know it failed and not why" unrecordable. A `COMPLETED` transition may carry failures
+ * for the same reason from the other side: a run that retried past a failed attempt and finished
+ * still happened.
+ *
+ * [fingerprint] is what a store compares a retry against; see [ExtractionRunFingerprint] for the
+ * encoding and why it is specified rather than left to a serializer.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property status The terminal state the run ends in
+ * @property finishedAt When the run reached it
+ * @property counts The run's final counts, or null to keep what the run already recorded
+ * @property failures The run's final failures, or null to keep what the run already recorded
+ */
+@ApiStatus.Experimental
+class ExtractionRunTransition @JvmOverloads constructor(
+    val status: ExtractionRunStatus,
+    val finishedAt: Instant,
+    val counts: ExtractionRunCounts? = null,
+    failures: List<ExtractionFailure>? = null,
+) {
+
+    /** The run's final failures, or null to keep what the run already recorded. */
+    val failures: List<ExtractionFailure>? =
+        failures?.let { Collections.unmodifiableList(ArrayList(it)) }
+
+    init {
+        require(status.isTerminal) {
+            "a transition ends a run, so its status must be terminal; $status is not"
+        }
+        val declared = this.failures
+        if (declared != null) {
+            require(declared.size <= ExtractionRunLimits.MAX_FAILURES) {
+                "a run may record at most ${ExtractionRunLimits.MAX_FAILURES} failures, " +
+                    "was ${declared.size}"
+            }
+        }
+    }
+
+    /**
+     * The digest a store compares a repeated terminal write against, computed once at construction.
+     */
+    val fingerprint: String =
+        ExtractionRunFingerprint.ofTerminal(status, finishedAt, counts, this.failures)
+
+    /**
+     * Derives the terminal run this transition produces from the running one it is applied to.
+     *
+     * Everything the transition does not own is carried across unchanged: lineage, tenant, start
+     * time, profile, source revisions, digests, runtime identity, requested configuration, subject
+     * references, experiment and cohort labels, replay fidelity, the invocation records the run
+     * accumulated, and its header version.
+     *
+     * @param run The run to terminalize. Must be [ExtractionRunStatus.RUNNING].
+     * @return The same run in its terminal state.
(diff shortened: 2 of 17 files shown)


dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt
```diff
@@ -0,0 +1,252 @@
...
+                )
+            }
+            // Invocation records are child rows, not header content. A save writes the header, and
+            // a caller updating counts from a run it loaded before an attempt was recorded must not
+            // delete that attempt. A durable backend gets this for free — a header MERGE never
+            // touches a child row — so the reference implementation has to do it explicitly or the
+            // two backends disagree.
+            val incoming = run.invocations.map { it.id }.toSet()
+            val merged = stored.invocations.filterNot { it.id in incoming } + run.invocations
+            val updated = rebuild(run, invocations = merged)
```
jimador: A stale header save can erase source revisions, counts, failures, profile, or fingerprints that another writer already stored because this rebuild takes every header field from the incoming run. Can we add versioning or merge the accumulating fields explicitly?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt
```diff
@@ -0,0 +1,252 @@
...
+    ): ExtractionRun {
+        synchronized(lock) {
+            val stored = runs[key] ?: throw ExtractionRunNotFoundException(key)
+            if (stored.status.isTerminal) {
+                throw ExtractionRunConflictException(
+                    key,
+                    "already ended as ${stored.status}; its invocation records are part of how it ended",
+                )
+            }
+            val merged = stored.invocations.filterNot { it.id == record.id } + record
```
jimador: A delayed `IN_FLIGHT` write can replace a terminal record for the same attempt here, which rewrites observed execution facts after success or failure. Can we make repeated writes idempotent and reject outcome regressions?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ProtectedContentRef.kt
```diff
@@ -0,0 +1,128 @@
...
+ * expired last week is a legitimate historical record.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property classification How restricted the material is
+ * @property expiresAt When the host's store stops holding the material
+ * @property handle The host's opaque identifier for it
+ */
+@ApiStatus.Experimental
+data class ProtectedContentRef(
```
jimador: Nothing attaches this to `ExtractionRun`, and DICE has no writer, reader, or retention behavior for it. Can we add this with the first runtime path that uses it?

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt
jimador: Header writes carry a version and take it under compare-and-set, so a stale save is rejected and can no longer overwrite source revisions, counts, failures, profile or fingerprints another writer stored. The contract suite pins each of those fields through both write doors, including clearing a stored value back to null.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt
jimador: Terminal records are locked. Repeating the same terminal payload succeeds as a replay; an incompatible rewrite is rejected on a fingerprint of the transition payload alone. SUCCEEDED, FAILED and CANCELLED lock alike, each pinned through both write doors.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ProtectedContentRef.kt
jimador: Removed from this PR. It returns with the first runtime path that writes, reads, or expires it.

jimador:
Invocation state has two write paths and no shared generation. Header CAS does not advance when `recordInvocation` writes a child. A stale header save with a non-empty snapshot can therefore erase a newer invocation, and concurrent `IN_FLIGHT` writers can silently lose facts.

That is not a safe audit store contract. Header updates should not rewrite invocation children. Give invocation records their own atomic insert/CAS semantics and test stale-header and concurrent-writer interleavings.

jimador:
Fixed. Each kind of state now has exactly one owner.

`save` writes header fields only. Whatever invocation list a caller hands it is carried through untouched, and an insert creates no invocation rows at all. `recordInvocation` is the sole door onto invocation state, insert-or-compare on the invocation's own key. A header save therefore cannot create, update or delete an invocation row, and a stale header carrying an old snapshot leaves a newer stored record alone.

I took this over a shared generation deliberately: making every invocation write advance the header would serialise an append-mostly audit path against itself. Run state accumulates from independent writers, the way a lineage event stream does — OpenLineage specifies that events describing one run are accumulative and no event rewrites an earlier one.

The contract changed with it. 14 cases that pinned header-owns-children were removed and 7 added, including the exact scenario you named — a stale header save carrying an old snapshot leaving the newer stored invocation intact — plus concurrent writers on separate attempts both landing. Every `recordInvocation` twin was kept unchanged.

Worth recording: one removed case was `a save and a recordInvocation racing on one attempt produce exactly one accepted terminal write`. That test was the defect made visible as a *passing* test, which is why thirteen adversarial gate rounds went past it — every round measured conformance to that assertion. Removing the merge machinery made the contract 150 lines smaller as well as correct.

Gate: `dice` 1476/0, `dice-storage` 309/0, contract suite 59/59 on both backends.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt
```diff
@@ -0,0 +1,349 @@
...
+ * **A header save is itself compare-and-set, on [ExtractionRun.version] — and it owns header
+ * fields only.** Two writers can hold the same running header at once, and whichever saves second
+ * must not silently put the other's write back the way it looked before. `save` accepts a write
+ * only when the version it names matches what is stored; a stale writer is told so, in an
+ * [ExtractionRunConflictException], and has to read the run again and rebuild its save on what it
+ * holds now. See `save`'s own KDoc for why this replaced an earlier, field-by-field merge of the
+ * running header. Whatever [run]'s `invocations` field carries plays no part in that comparison or
+ * in what gets stored — see `save`'s own KDoc for why.
+ *
+ * **`recordInvocation` is the only door onto invocation rows, and each row keeps its own
```
jimador, line 46: Invocation state had two write paths and no shared generation. Header CAS did not advance when `recordInvocation` wrote a child. A stale header save with a non-empty snapshot could therefore erase a newer invocation, and concurrent `IN_FLIGHT` writers could silently lose facts.

That is not a safe audit store contract. Header updates should not rewrite invocation children. Give invocation records their own atomic insert/CAS semantics and test stale-header and concurrent-writer interleavings.

jimador: Fixed. Each kind of state now has exactly one owner.

`save` writes header fields only. Whatever invocation list a caller hands it is carried through untouched, and an insert creates no invocation rows at all. `recordInvocation` is the sole door onto invocation state, insert-or-compare on the invocation's own key. A header save therefore cannot create, update or delete an invocation row, and a stale header carrying an old snapshot leaves a newer stored record alone.

I took this over a shared generation deliberately: making every invocation write advance the header would serialise an append-mostly audit path against itself. Run state accumulates from independent writers, the way a lineage event stream does; OpenLineage specifies that events describing one run are accumulative and no event rewrites an earlier one.

The contract changed with it. 14 cases that pinned header-owns-children were removed and 7 added, including the exact scenario you named, a stale header save carrying an old snapshot leaving the newer stored invocation intact, plus concurrent writers on separate attempts both landing. Every `recordInvocation` twin was kept unchanged.

Worth recording: one removed case was `a save and a recordInvocation racing on one attempt produce exactly one accepted terminal write`. That test was the defect made visible as a *passing* test, which is why thirteen adversarial gate rounds went past it. Removing the merge machinery made the contract 150 lines smaller as well as correct.

Gate: `dice` 1476/0, `dice-storage` 309/0, contract suite 59/59 on both backends.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/ExtractionRunTransition.kt
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    /**
+     * The digest a store compares a repeated terminal write against, computed once at construction.
+     *
+     * It covers this transition's identity — [status] and [finishedAt] — and nothing else. See
+     * [ExtractionRunFingerprint] for why [counts] and [failures] travel as data beside it.
+     */
+    val fingerprint: String = ExtractionRunFingerprint.ofTerminal(status, finishedAt)
```
jimador, line 85: Terminal fingerprint versus #69. `ExtractionRunTransition.fingerprint` hashed `counts` and `failures`, and `ExtractionRunCounts` is five plain ints. #69's typed product outcomes change that payload, which changes a persisted fingerprint format. Either hash only the transition identity (status, finishedAt) and carry counts as data, or plan on `xrun-terminal:v2` before this ships in a release. State which in the design doc.

jimador: **Hash transition identity only.** Landed as "Hash only transition identity and announce applied transitions". `ofTerminal` covers `(status, finishedAt)`; counts and failures travel as data on the transition and still participate in its equality. `TERMINAL_VERSION` moved to `xrun-terminal:v2` once, before any release, and this doc states the choice: #69's typed product outcomes change no persisted fingerprint format.

dice/src/main/kotlin/com/embabel/dice/common/DiceEvent.kt
```diff
@@ -160,6 +162,30 @@ data class ExtractionBatchCompleted @JvmOverloads constructor(
...
+ * is no source text or provider message anywhere in it, so this event is safe to hand to a listener
+ * that logs or forwards what it receives.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ *
+ * @property run The run in its terminal state.
+ * @property timestamp When the event was created.
+ */
+@ApiStatus.Experimental
+data class ExtractionRunTransitioned @JvmOverloads constructor(
```
jimador, line 184: No `DiceEvent` was emitted for any run transition, so the Wave F audit projection and any host listener had nothing to subscribe to. And `ExtractionRunFingerprint.ofFields` with `HEADER_VERSION` and `INVOCATION_VERSION` had no caller because `DrivineExtractionRunStore` compared in Cypher. Drop it or use it.

jimador: `ExtractionRunTransitioned` is a `DiceEvent`: both stores emit it after an applied terminal transition, gated on the result, nothing wired automatically, so the audit projection and host listeners have their subscription point. `ofFields` is used: the Drivine row mappers digest header and invocation rows through it under `HEADER_VERSION` and `INVOCATION_VERSION`, backing the versioned compare-and-set header writes.

igordayen:
from CODEX:

Findings:

  - P2 InMemoryExtractionRunStore retains every run forever
    dice/src/main/kotlin/com/embabel/dice/proposition/extraction/InMemoryExtractionRunStore.kt:79-82 keeps all run headers and terminal
    fingerprints in maps, and ExtractionRunStore exposes no delete/prune/retention API. Bounded reads only bound the returned page, not
    stored memory.
    Example: a host using the shipped in-memory store for 50k extractions/day keeps every completed run and fingerprint until JVM
    restart. Either add explicit retention/pruning support, cap the in-memory implementation, or make it clearly dev/test-only so a
    long-running host cannot accidentally choose an unbounded audit store.

  - P2 reads become global stop-the-world scans as the store grows
    InMemoryExtractionRunStore.page() filters and sorts runs.values while holding the single monitor (:260-267). runsInContext,
    childrenOf, and runsOfRoot all use that path. The API returns at most limit, but the work is still O(total stored runs), across all
    tenants, under the same lock used by save, recordInvocation, and transition.
    Example: after millions of historical runs, runsInContext("tenant-a", 10, since = null) scans/sorts the whole map and blocks all
    writers while it does so. Tenant-indexed maps or per-query indexes would keep the reference store’s thread-safety claim from
    degrading under retention.

  - P3 listener failures can make a committed transition look failed to the caller
    transition() commits the terminal run, releases the lock, then calls listener.onEvent(...) without catching
    (InMemoryExtractionRunStore.kt:178-203). If the listener throws, the run is already terminal, but the caller sees an exception;
    retrying the same transition returns REPLAYED and emits no event.
    SafeDiceEventListener exists, but the store does not wrap the collaborator itself. If “transition succeeded” must not depend on
    listener quality, wrap internally or document this as a hard contract on callers.

igordayen (commented):
@jimador - would you please see few inquiries from myself and CODEX. Do we need to consider eviction policy to control memory growth?
Thanks

jimador:
All three addressed.

- Retention: `InMemoryExtractionRunStore` takes `maxRuns` (default 10_000) and evicts the oldest ended runs past it, never a running one; the class doc says what a host running the reference store in production is accepting. e5be31b9.
- Reads: each tenant's runs sit in their own index, so a page reads that tenant only; filter, order, limit is unchanged. b11703a6.
- Listener: a throwing listener is logged at error and the transition stays committed and reports `APPLIED`. 84cbd605.

On your eviction question: yes for the reference store, and the cap above is it. For the Drivine store, retention past that is the durable store's own policy, kept for as long as an operator decides, and a sweep on `finishedAt` would be a separate slice; the design doc now says so.

igordayen (approved):
@jimador - looks good subject to rebasing, thanks

PR #99, opened 2026-08-31
Extraction runs: the Drivine store

jimador:
PR 8 of the extraction-integrity train (refs #67), stacked on #98. The Drivine/Neo4j run store. The terminal transition is one Cypher statement taking the node's exclusive lock before reading status, backstopped by a uniqueness-constrained terminal-write node, so two writers that both saw RUNNING cannot both commit; the constraint's sufficiency is proven by mutation (removing both guards yields six racing writers, all APPLIED, three contradictory endings for one run), the lock's by the documented isolation argument, and the docs say which is which. Fingerprint strings are stored verbatim, compared verbatim, never re-derived — the inherited contract case that kills re-derivation passes. Invocation records live as child rows on their own constrained key `(contextId, runId, index, attempt)` and survive header saves; every read is tenant-scoped ahead of its LIMIT and cross-tenant fail-closed is integration-proven; `ContextId` is capped at 1024 on writes, closing the run key's last unbounded component. The #98 contract suite runs unmodified against this store, plus Drivine-specific integration tests including the multi-process-shaped race.

- **The failure vocabulary persists as typed columns.** Failures store `code`, `stage`, `providerStatus`, `measure`, `at` and the invocation reference; no stored property is shaped to hold text, so the model-level privacy guarantee survives the round-trip.
- **Applied transitions are announced from Drivine.** The store takes a constructor-injected `DiceEventListener` and emits `ExtractionRunTransitioned` exactly once per run that ended, matching the in-memory reference; the base suite's event cases run inherited against Neo4j.
- Header and invocation rows digest through `ExtractionRunFingerprint.ofFields` under `HEADER_VERSION` and `INVOCATION_VERSION`, backing the versioned compare-and-set header writes from #98.

**Breaking changes: none.** New classes in `dice-storage`; hosts adopting the run store declare the schema items the CHANGELOG enumerates.

**Opt-in and status:** **EXPERIMENTAL.** `DrivineExtractionRunStore` activates only where a host wires it; nothing in DICE produces a run yet — the coordinator is the first real caller and is sequenced next, after #101's lineage write.

**Next in stack:** #101 — proposition lineage, and the canonical-persistence rule.

### DIFF
--- a/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
+++ b/dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
@@ -1,0 +1,981 @@
+/*
+ * Copyright 2024-2026 Embabel Pty Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.embabel.dice.storage
+
+import com.embabel.dice.proposition.extraction.ExtractionInvocationOutcome
+import com.embabel.dice.proposition.extraction.ExtractionInvocationRecord
+import com.embabel.dice.proposition.extraction.ExtractionRun
+import com.embabel.dice.proposition.extraction.ExtractionRunConflictException
+import com.embabel.dice.proposition.extraction.ExtractionRunKey
+import com.embabel.dice.proposition.extraction.ExtractionRunNotFoundException
+import com.embabel.dice.proposition.extraction.ExtractionRunRef
+import com.embabel.dice.proposition.extraction.ExtractionRunStatus
+import com.embabel.dice.proposition.extraction.ExtractionRunStore
+import com.embabel.dice.proposition.extraction.ExtractionRunTransition
+import com.embabel.dice.proposition.extraction.ExtractionRunTransitionOutcome
+import com.embabel.dice.proposition.extraction.ExtractionRunTransitionResult
+import org.drivine.manager.PersistenceManager
+import org.drivine.query.QuerySpecification
+import org.slf4j.LoggerFactory
+import org.springframework.transaction.PlatformTransactionManager
+import org.springframework.transaction.annotation.Propagation
+import org.springframework.transaction.annotation.Transactional
+import org.springframework.transaction.support.TransactionSynchronizationManager
+import org.springframework.transaction.support.TransactionTemplate
+import java.time.Clock
+import java.time.Instant
+import java.util.UUID
+
+/**
+ * Drivine / Neo4j implementation of [ExtractionRunStore].
+ *
+ * ## The graph
+ *
+ * - `(:ExtractionRun {contextId, runId, ...})` — the run header, one node per run per tenant.
+ * - `(:ExtractionRun)-[:RECORDED]->(:ExtractionRunInvocation {contextId, runId, invocationIndex, attempt, ...})`
+ *   — one child node per attempt at one planned model call.
+ * - `(:ExtractionRun)-[:ENDED_BY]->(:ExtractionRunTerminalWrite {contextId, runId, fingerprint, ...})`
+ *   — at most one per run, ever, and the thing that makes that "at most one" true.
+ *
+ * Every key is tenant-qualified because a run id is host-minted and DICE never assumes it is
+ * globally unique. Unlike `DrivineDriftReportStore`, the tenant needs no `ctx:`-prefixed stand-in:
+ * that store's scope is nullable, and a Cypher MERGE cannot key on a null, so a global report needed
+ * a non-null encoding that no real context id could collide with. A run's tenant is never null, so
+ * the plain value is already an injective key.
+ *
+ * Every statement is parameterized; nothing caller-derived is interpolated into Cypher. The
+ * statements are assembled only from this class's own literals.
+ *
+ * See [ExtractionRunSchema] for the constraints and indexes this depends on. They are not tuning:
+ * a MERGE on a natural key is race-free only under a uniqueness constraint on that key.
+ *
+ * ## Compare-and-set: why a terminal write cannot be applied twice
+ *
+ * The whole of [transition] is one Cypher statement, so it is one transaction, and it works two
+ * ways at once — a lock that makes the race rare and a constraint that makes the wrong answer
+ * impossible.
+ *
+ * **The lock.** The statement's first act after matching the run is `SET n.casLock = $lockToken`,
+ * before it reads anything. A property write takes an exclusive lock on the node and holds it until
+ * commit, so a second transaction reaching that line blocks until the first commits. Neo4j reads at
+ * read-committed, and the read of `n.status` is downstream of the `SET` in the same statement, so
+ * the second transaction reads the status *after* it acquired the lock and therefore *after* the
+ * first transaction committed. It sees a terminal run and takes the no-op branch. This is the same
+ * write-lock-before-read idiom `DrivineMetamodelVersionStore` uses on its counter.
+ *
+ * Two details make that argument hold rather than nearly hold. The lock token is a fresh UUID on
+ * every call, so the write is always a real change and never a no-op a future Neo4j might optimize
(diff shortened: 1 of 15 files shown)


jimador:
The Neo4j implementation has three transaction and replay defects:

- A mixed invocation batch writes accepted rows before throwing on a conflict. Because the exception is caught inside the ambient transaction, those partial writes can commit.
- The transient-conflict retry reruns a statement inside the same transaction that has already failed; it cannot provide a clean retry boundary.
- The durable header and invocation fingerprint is raw JSON serialization, so serializer/configuration changes can turn equivalent state into a conflict. The canonical versioned codec exists to avoid this.

There are related write-then-throw paths that can leave stubs or orphaned state. Make validation precede mutation, move retries outside the transaction boundary, and fingerprint canonical bytes. Cover all of this with ambient-transaction tests.

jimador:
All three fixed, plus the store reworked to the corrected #98 contract.

**Partial commits.** The original shape is gone: `save` no longer writes invocations at all, and `recordInvocation` was never a batched statement. What remains is the same shape one call level up — several `recordInvocation` calls sharing one caller-owned transaction, one of which conflicts. Each invocation row is independently keyed and settles on its own merits: the accepted writes commit, and the conflicting write leaves the row it targeted byte-identical to what was there. Covered by an ambient-transaction test asserting the locked record is unchanged, and by a second proving a rejected write leaves no orphan node even when the transaction that caught it commits.

**Retry boundary.** Two concurrent MERGE-creates of the same not-yet-existing invocation node can trip Neo4j's deadlock detector, which aborts a transaction outright. Retrying inside a transaction the database already discarded is a retry against nothing. A recovered writer now reads and writes in a transaction the failure never touched, proven by a five-round race test that would hang or leave two disagreeing writes if it were retrying inside the dead one.

**Fingerprints.** Header and invocation digests come off the canonical length-prefixed encoding with a version tag per kind. Bind-map ordering and `Map` implementation no longer move a digest, absent is distinguished from empty string, and a header digest cannot collide with an invocation or terminal one over the same field set. Stored verbatim, compared verbatim, never re-derived.

Gate: `dice` 1476/0, `dice-storage` 309/0. The #98 contract suite runs unmodified against this store, 59/59, matching the in-memory reference exactly.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
```diff
@@ -0,0 +1,1031 @@
...
+ * **The fingerprint is stored, never re-derived.** The node carries the exact string
+ * [ExtractionRunTransition.fingerprint] computed, and a repeated terminal write is decided by
+ * comparing its fingerprint against that string. Re-deriving a digest from the stored run would
+ * reject a correct retry whenever an attempt had been recorded in between, because the run it
+ * derived from would have changed while the terminal write did not.
+ *
+ * ## A header write never touches a child row
+ *
+ * Invocation records are their own nodes on their own key, and [save]'s Cypher names none of them:
+ * it reads and writes `(:ExtractionRun)` alone. [recordInvocation] is the only method that ever
```
jimador, line 111: A mixed invocation batch wrote accepted rows before throwing on a conflict. Because the exception was caught inside the ambient transaction, those partial writes could commit. There are related write-then-throw paths that can leave stubs or orphaned state. Make validation precede mutation and cover it with ambient-transaction tests.

jimador: **Partial commits.** The original shape is gone: `save` no longer writes invocations at all, and `recordInvocation` was never a batched statement. What remains is the same shape one call level up: several `recordInvocation` calls sharing one caller-owned transaction, one of which conflicts. Each invocation row is independently keyed and settles on its own merits: the accepted writes commit, and the conflicting write leaves the row it targeted byte-identical to what was there. Covered by an ambient-transaction test asserting the locked record is unchanged, and by a second proving a rejected write leaves no orphan node even when the transaction that caught it commits.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
```diff
@@ -0,0 +1,1031 @@
...
+ * terminal write per run is a schema fact, not an inference.
+ *
+ * That is the multi-process half of the guarantee. Threads in one JVM could be serialized by a
+ * monitor; two processes cannot be, and nothing in this class holds mutable state to serialize on.
+ * Both halves come entirely from the database.
+ *
+ * **The fingerprint is stored, never re-derived.** The node carries the exact string
+ * [ExtractionRunTransition.fingerprint] computed, and a repeated terminal write is decided by
+ * comparing its fingerprint against that string. Re-deriving a digest from the stored run would
+ * reject a correct retry whenever an attempt had been recorded in between, because the run it
```
jimador, line 105: The transient-conflict retry reran a statement inside the same transaction that had already failed; it cannot provide a clean retry boundary. Move retries outside the transaction boundary.

jimador: **Retry boundary.** Two concurrent MERGE-creates of the same not-yet-existing invocation node can trip Neo4j's deadlock detector, which aborts a transaction outright. Retrying inside a transaction the database already discarded is a retry against nothing. A recovered writer now reads and writes in a transaction the failure never touched, proven by a five-round race test that would hang or leave two disagreeing writes if it were retrying inside the dead one.

dice-storage/src/main/kotlin/com/embabel/dice/storage/ExtractionRunRowMappers.kt
```diff
@@ -0,0 +1,668 @@
...
+     *
+     * Digested through [ExtractionRunFingerprint.ofFields], the canonical codec. A serializer's
+     * output can move without the header moving with it — a different `Map` implementation
+     * iterating [header] in a different order, a library upgrade changing how it renders a number —
+     * and every one of those would read as a genuine change to a save that is really a no-op or a
+     * byte-identical retry. The canonical encoding sorts by field name and renders every value as
+     * the plain string it already is, so the digest is a function of the header's content alone.
+     */
+    fun headerFingerprint(header: Map<String, Any?>): String = ExtractionRunFingerprint.ofFields(
+        ExtractionRunFingerprint.HEADER_VERSION,
```
jimador, line 184: The durable header and invocation fingerprint was raw JSON serialization, so serializer/configuration changes could turn equivalent state into a conflict. The canonical versioned codec exists to avoid this. Fingerprint canonical bytes.

jimador: **Fingerprints.** Header and invocation digests come off the canonical length-prefixed encoding with a version tag per kind, through `ExtractionRunFingerprint.ofFields` under `HEADER_VERSION` and `INVOCATION_VERSION`. Bind-map ordering and `Map` implementation no longer move a digest, absent is distinguished from empty string, and a header digest cannot collide with an invocation or terminal one over the same field set. Stored verbatim, compared verbatim, never re-derived.

Gate: `dice` 1476/0, `dice-storage` 309/0. The #98 contract suite runs unmodified against this store, 59/59, matching the in-memory reference exactly.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
```diff
@@ -0,0 +1,1031 @@
...
+ * A false positive is bounded here: the caller only ever treats it as "someone else recorded the
+ * terminal write", and re-reads to find out. If there is no terminal write, the original exception is
+ * rethrown.
+ */
+private fun isUniquenessViolation(error: Throwable?): Boolean {
+    var current: Throwable? = error
+    val seen = mutableSetOf<Throwable>()
+    while (current != null && seen.add(current)) {
+        val message = current.message ?: ""
+        if (message.contains("ConstraintValidationFailed", ignoreCase = true) ||
```
igordayen: what if message changes?


does not spring data have portable exception equivalent? (ConstraintViolationException?)

igordayen (commented):
@jimador - looks good, could you please review few inquiries i left, thank you

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivineExtractionRunStore.kt
jimador: Right, the message is not a contract. The driver's exception carries the Neo4j status code, and a uniqueness violation is always `Neo.ClientError.Schema.ConstraintValidationFailed`, so both stores now check that through one `Neo4jErrors` helper and match no text at all (d219ce78, 0a0149bd). Drivine does not translate into Spring's `DataAccessException` family, so the status code is the portable answer here. The race integration tests still pass against a real Neo4j, which is the proof the code is what the server sends.

igordayen (approved):
@jimador - looks good, subject to rebasing, thanks

Issue #68, opened 2026-07-24
Atomically commit extraction outcomes and disposition audit


jimador:
## Goal

Atomically persist extraction-run outcomes, disposition audit, and accepted propositions
after the host supplies its policy decisions.

## Context

The pipeline returns unsaved propositions, while legacy convenience paths persist
internally. The new path needs one DICE storage contract so a host crash cannot leave a
completed run without products or committed products without their run and disposition.

## Acceptance criteria

- [ ] The commit input contains run state, typed product outcomes, host disposition
      decisions, accepted products, review records, and rejection records.
- [ ] One transaction persists accepted propositions, run links, outcomes, and audit.
- [ ] Retrying the same commit is idempotent.
- [ ] Failed commits leave the run retryable and do not complete its analysis claim.
- [ ] Partial product success is preserved without committing failed products.
- [ ] Review, reject, demote, and projection-exclusion decisions remain queryable.
- [ ] Drivine integration tests cover crash boundaries, retry, partial success, and tenants.

## Constraints

- Depends on durable extraction runs.
- The host application composes policy; DICE owns persistence and transaction execution.
- Legacy persistence remains available during migration.

## Out of scope

- Host policy implementation.
- Source revision detection.
- Exactly-once delivery outside the DICE store transaction.

## Original request

Close the extraction-to-persistence gap with a durable commit boundary that preserves
successful work, audit decisions, and retry safety.

PR #101, opened 2026-08-31
Extraction runs: proposition lineage

jimador:
Slice 9 of the extraction-integrity train (#67), stacked on #99. A stored claim can now be traced to the run that produced it. `PropositionStore.saveAllReturningCanonical` and `PersistablePropositions.persistReturningCanonical` are additive default-bodied calls returning `PropositionPersistenceResult` — the canonical proposition for each input, the input-id to stored-id mapping, and a distinct-by-canonical-id view; the result type rejects length mismatches, one input landing on two ids, per-position tenant mismatches, and one canonical id answered under two contexts. `PropositionRunLinkStore` (in-memory and Drivine, behind a 14-case cross-backend contract suite) writes a `MERGE`-idempotent `(:Proposition)-[:PRODUCED_BY_RUN]->(:ExtractionRun)` edge, tenant-guarded on both endpoints, with all-or-nothing batch writes proven inside the TOCTOU window. The headline invariant on Neo4j: two identical-content runs leave one proposition, one source grounding, and two run links. Lineage is recorded immediately after the claims are saved, before any fallible pass, so a throwing projector or grounding service leaves saved claims attributed and still propagates.

- **Canonical persistence is the only path.** `persistAndProject` used to take the repository's canonical propositions only when a run was present and the pre-save minted ones otherwise, which let an audit setting decide whether the graph was written correctly. It now wires projection, grounding and structural relationships against what `save` returned on every call; a run's only effect is the lineage write. A mutation restoring the old no-run wiring fails three tests, projection-targets-a-pre-save-id among them.
- **Attribution fails loud.** A run present with no way to record its lineage is an error by policy; duplicate `PRODUCED_BY_RUN` edges cannot occur (`MERGE` on the pair, proven by the linking-twice case).
- `ExtractionRun.invocations` normalize to plan order at construction (the #99 follow-up); stored fingerprints are unaffected.
- The experimental extraction surfaces are marked as such in the CHANGELOG.

**Breaking changes: none.** Every existing public signature, descriptor and default-body behavior is byte-identical, pinned by a binary-compatibility test asserting the `@JvmOverloads` synthetic constructor descriptors (arities 9–16) survived. **One behavioral change, sanctioned and on an unreleased experimental surface:** the no-run extraction path now also persists, projects and grounds against canonical ids — the audit-metadata-never-changes-product-behavior rule, applied.

**Opt-in and status:** **EXPERIMENTAL**, marked `@ApiStatus.Experimental` in code. The Drivine stores and the run schema DDL sit behind `embabel.dice.extraction.runs.enabled` (default false) alongside the graph backend condition; a host that never sets it gets no run beans and no constraints or indexes written, and the CHANGELOG names exactly what turning it on writes. Run attribution happens when `ExtractionRequest.currentRun` is set and a `PropositionRunLinkStore` is bound; with neither, behavior is the canonical path alone. Honest limit, in shipped docs: lineage is best-effort for application-level failures only — a statement failing at the server inside a wrapping host transaction terminates that transaction beneath Spring, demonstrated by a fault-injection test; the structural fix (committing claims before recording lineage) belongs to the coordinator slice and the docs name it.

**The lineage shape.** The root ref is denormalized onto every run, so a whole-lineage lookup is one indexed read, with no chain walk:

```mermaid
flowchart LR
    R[run A, pass 0, root = self] --> B[run B, pass 1, parent = A]
    B --> C[run C, pass 2, parent = B]
    B -. "root = A" .-> R
    C -. "root = A" .-> R
    PA[proposition] -->|PRODUCED_BY_RUN| C
    PB[proposition] -->|PRODUCED_BY_RUN| C
```

Chain walks stay bounded and cycle-safe, because the denormalized root is a read optimization and never the integrity check.

## Landing

- Build the union of this stack's tip and the metamodel stack's tip (#88) before either lands; the two share six recurring conflict files with recorded resolutions; `DiceStorageAutoConfiguration.kt` conflicts too and is resolved by hand from the union worktree's committed adaptation, which also carries the runs gate on both schema beans.
- Whichever stack lands second restacks and applies one adaptation: `ExtractionRunSchema` joins the metamodel stack's `DiceStorageSchema` derived-ownership contract (object declaration, `override` on `specs()`, bookkeeping relationship types, catalog and test beans). `DiceStorageSchemaRegistrationTest` fails until it is applied — that is the guard working.
- No consumer migration is required to land: the four-argument `undoSingleCollapse` stays as a deprecated overload (#92). Downstream callers move to `CollapseUndoCommand` when convenient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1724,7 +1724,7 @@ and the consumer PRs that deliver it).
 
 - **EXPERIMENTAL.** Extraction-run lineage: a stored claim can now be traced to the runs that
   produced it, and the canonical id it was stored under is no longer thrown away. Two halves.
-  **The persistence-result seam.** `DrivinePropositionRepository.save` has always answered a fresh
+  **Canonical ids come back from a save.** `DrivinePropositionRepository.save` has always answered a fresh
   insert of text it already holds with the proposition it already holds — a different id from the one
   extraction minted — and `PropositionStore.saveAll` returns `Unit`, so callers never learned it. Any
   edge, projection or grounding link written afterwards against the minted id points at a node that
@@ -1740,8 +1740,8 @@ and the consumer PRs that deliver it).
   wiring, projection and grounding run over, so inputs that deduplicated together are one unit of
   downstream work rather than one per input, which would repeat idempotent edge writes and inflate
   the records written about them. `of` also rejects a canonical result carrying a different
-  `ContextId` from its input: lineage would refuse to link a foreign-tenant proposition, but lineage
-  is best-effort and refuses quietly, so the check belongs where the object enters the pipeline. It
+  `ContextId` from its input: a foreign-tenant proposition has no business reaching a link at all,
+  so the check belongs where the object enters the pipeline. It
   also rejects one stored id answered under two different contexts, which each per-position check
   would pass individually while the resolution step handed the earlier position the other tenant's
   object. When
@@ -1802,24 +1802,47 @@ and the consumer PRs that deliver it).
   `@JvmOverloads` arity, and that no constructor mentions the lineage store at all. The binding is
   **one-time**: a second `withRunLineage` call throws `IllegalStateException` rather than silently
   swapping or clearing the store an in-flight extraction is about to record against.
-  **Behavioural, run-present flows only:** when a `SourceAnalysisContext.currentRun` is present,
-  `persistAndProject` now wires structural relationships, graph projection and grounding against the
-  propositions the repository returned rather than the ones extraction minted, and writes the run
-  links. An analysis with no run takes the previous path unchanged, pre-save objects included, and
-  two tests pin that so "unchanged" fails if it stops being true. The switch is the run, not the
-  presence of a link store — a host that passes `currentRun` opted into #67's experimental surface in
-  the profiles slice, and gets correct edges whether or not it records lineage. Lineage is written
-  best-effort and **directly behind the save, ahead of structural wiring, projection and
-  grounding**: the claims are saved at that point and no fallible pass has run yet. That needed
-  `persistReturningCanonical` split into `persistCanonicalPropositions` and
-  `wireStructuralRelationships` — both published, the original preserved as their composition — so
-  lineage can run between them; structural wiring is the first fallible pass and used to sit inside
-  the save. Running attribution after any of them meant a throw could leave stored claims with no
-  record of the run that produced them — the one outcome the relation exists to
-  prevent, arriving exactly when the audit matters most. A link that cannot be written is logged
-  with its exception and the extraction stands.
-  **What best-effort covers, exactly.** The lineage write joins a caller's transaction rather than
-  opening its own: `REQUIRES_NEW` would suspend that transaction, and a suspended transaction's
+  **Compatibility: behavioral, every extraction flow — a fix.** `persistAndProject` now wires
+  structural relationships, graph projection and grounding against the propositions the repository
+  returned, on every call, with or without a run. It used to do that only when a
+  `SourceAnalysisContext.currentRun` was present and to use the pre-save objects otherwise, which
+  made an audit setting decide whether the graph was written correctly: a host that turned on
+  extraction runs silently got different edges, and a host that did not kept writing edges against
+  ids the store does not hold. Under dedup or merge those phantom pre-save ids reached projection and
+  grounding, so the edges pointed at nodes that were never stored — the same class of defect the
+  `POST /extract` response fix above closed, arriving here in the write path. Audit
+  metadata never changes product behaviour: a run now adds a lineage write and nothing else, and a
+  test runs the same extraction with and without one and compares everything except that write.
+  Hosts that never passed a run get corrected edges under dedup without changing a line. Lineage is
+  written **last, after structural wiring, projection and grounding have all completed**, so that a
+  failure it raises leaves a complete extraction behind: claims persisted, structural edges wired,
+  projection run, grounding run, and no `PRODUCED_BY_RUN` edge. That is the end state a `STRICT`
+  failure reports, and `LENIENT` reaches the same one and reports success. The split of
+  `persistReturningCanonical` into `persistCanonicalPropositions` and `wireStructuralRelationships`
+  stays — both published, the original preserved as their composition — because the canonical
+  propositions a save returns are what every later pass wires against.
+  An earlier cut wrote lineage directly behind the save, ahead of the three wiring passes, so a
+  throwing projector could not leave stored claims unattributed. That ordering cannot survive
+  failing loud: raising from behind the save returns through the middle of the pipeline with the
+  claims stored and projection and grounding silently skipped, which is a partial state nothing
+  declared. Attribution is a statement about finished work, so it is made when the work is finished;
+  the accepted trade is that a pass throwing before lineage means no attribution is written, and the
+  honest report of that is a failed extraction with no run edge.
+  **Attribution fails loud by policy.** A new `LineageFailurePolicy` says what happens when lineage
+  cannot be written, and `STRICT` is the default. Under it, two things fail the extraction: an
+  analysis carrying a run with no `PropositionRunLinkStore` bound, and a link write that throws. Both
+  used to be swallowed — the missing store returned quietly and every `RuntimeException` went to a
+  log line — so the two failures an operator most needs to hear about, lineage wired wrong and
+  lineage refusing writes, both reported success. An audit surface whose absence is silent is worth
+  nothing at the one moment it is consulted. `LENIENT` restores the old behaviour for a host that has
+  decided the claims outweigh their attribution and says so in configuration; the failure carries its
+  cause either way, so a scope rejection and an outage stay distinguishable. Bound with the store:
+  `withRunLineage(store, policy)` is a second overload, for the same descriptor reason the
+  one-argument form exists at all. Failures raise
+  `LineageNotRecordedException`. An analysis that saved nothing records nothing and fails under
+  neither policy.
+  **What joining a caller's transaction covers, exactly.** The lineage write joins a caller's
+  transaction and never opens its own: `REQUIRES_NEW` would suspend that transaction, and a suspended transaction's
   uncommitted propositions are invisible, so a host wrapping extraction in `@Transactional` would get
   fail-closed lineage on every extraction. Joining means Spring marks the participating transaction
   rollback-only when `link` throws — Drivine overrides `doSetRollbackOnly` and the flag is set — but
@@ -1828,7 +1851,9 @@ and the consumer PRs that deliver it).
   pinned by a test that goes red if either changes. The guarantee therefore covers the failures
   `link` raises itself, all of which are thrown after its statements succeeded. It does **not** cover
   a statement that fails at the server, which terminates the transaction beneath Spring where no
-  catch reaches; a test injects that and measures the cost. Hosts that do not wrap extraction in a
+  catch reaches; a test injects that and measures the cost. Under `STRICT` a raised failure reaches
+  the caller, so inside a host's `@Transactional` the claims and their lineage roll back together —
+  which is what strict attribution asks for. Hosts that do not wrap extraction in a
   transaction are unaffected, since each save has already committed. DICE #67 slice 10 closes the
   window by committing claims before recording lineage.
   **Behavioural for equality, on an unreleased type:** `ExtractionRun.invocations` is now normalized
@@ -1842,9 +1867,22 @@ and the consumer PRs that deliver it).
   because a caller comparing runs or reading `invocations[0]` would see it.
   **No new schema and no migration.** `ExtractionRunSchema.specs()` is unchanged: both endpoint
   labels already carry the uniqueness constraints these statements seek on, and a relationship has no
-  key of its own, since `MERGE` on a pattern between two matched nodes creates at most one edge.
-  `ExtractionRunSchema` gains a `PRODUCED_BY_RUN_REL` constant. Nothing is auto-configured, so a host
-  opts in by declaring `DrivinePropositionRunLinkStore` and passing it to
-  `IncrementalPropositionExtraction`. No released DICE ever wrote this relationship type. Every new
+  key of its own, since `MERGE` on a pattern between two matched nodes creates at most one edge and
+  Neo4j locks the endpoints when it decides to create. Duplicate `PRODUCED_BY_RUN` edges are pinned
+  from both sides: the in-memory store keeps the relation as a set behind one monitor and a
+  concurrency test drives eight threads at one pair, and the Drivine store counts the relationships
+  themselves after linking the same pair twice, because every read in the contract returns refs and
+  ids, so a duplicate edge is invisible to all of them. A relationship-level uniqueness
+  constraint is not available to back this up: Drivine's schema vocabulary is node-scoped
+  (`UniquenessConstraintSpec` takes a label), so there is nothing to declare in a `SchemaCatalog`.
+  `ExtractionRunSchema` gains a `PRODUCED_BY_RUN_REL` constant.
+  **Now auto-configured, on the graph backend, inert without a run.** `dice-storage-autoconfigure`
+  registers `DrivineExtractionRunStore`, `DrivinePropositionRunLinkStore` and the run schema catalog
+  under the same `embabel.dice.store.type=graph` condition and the same `@ConditionalOnMissingBean`
+  posture as every store beside them; they used to be declared only by `dice-storage`'s own
+  `TestApplication`, so the suite exercised them and no host could get them without writing the beans
+  by hand. Registering them changes nothing on its own — both are unreachable until a caller names a
+  run on an `ExtractionRequest`, and lineage additionally has to be bound with `withRunLineage` — and
+  a host that declares its own keeps them. No released DICE ever wrote this relationship type. Every new
   type carries `@ApiStatus.Experimental` and the shapes may still move while the remaining #67 slices
   land.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -642,7 +642,7 @@ and the consumer PRs that deliver it).
   application with no `DeclaredSchemaSource` bean sees no change. Nothing appears on an application's
   HTTP surface until it imports `DiceRestConfiguration`.
 
-- Optional source revisions in the `dice` core provenance model, the first slice of DICE #64.
+- Optional source revisions in the `dice` core provenance model, the first slice of DICE #64. **EXPERIMENTAL** (shape may change before 1.0) — opt-in: a store implements `SourceRevisionQueryCapable`.
   `ProvenanceEntry` gains a sixth field, `sourceRevision`: an opaque, provider-defined string,
   non-blank when present, recording which version of a source a claim was read from.
   `SourceLocator.key()` is untouched, so one document read at two revisions is still one source
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1844,6 +1844,14 @@ and the consumer PRs that deliver it).
   one-argument form exists at all. Failures raise
   `LineageNotRecordedException`. An analysis that saved nothing records nothing and fails under
   neither policy.
+  **The async event path is as loud as a direct call** (PR #101 review). `extractPropositions`
+  used to swallow every exception from an event-published extraction, so a `LineageNotRecordedException`
+  under `STRICT` was logged at `warn` and lost while the same inputs on a direct call failed. It is
+  now logged at `error` with the run key and rethrown to the publisher; the queue drain keeps going
+  past a failing event and raises the first failure once the queue is empty. **Compatibility:
+  behavioral.** A host publishing `SourceAnalysisRequestEvent`s with a run and `STRICT` lineage
+  now sees the failure where its event multicaster reports listener exceptions; under `LENIENT`
+  nothing changes.
   **What joining a caller's transaction covers, exactly.** The lineage write joins a caller's
   transaction and never opens its own: `REQUIRES_NEW` would suspend that transaction, and a suspended transaction's
   uncommitted propositions are invisible, so a host wrapping extraction in `@Transactional` would get
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1895,6 +1895,10 @@ and the consumer PRs that deliver it).
   `dice-storage`'s own `TestApplication`, so the suite exercised them and no host could get them
   without writing the beans by hand; a host that declares its own keeps them, under the same
   `@ConditionalOnMissingBean` posture as every store beside them.
+  **The in-memory link store prunes what it can see is gone** (PR #101 review). A read that finds a
+  link whose proposition is gone altogether drops the id, and an emptied run
+  entry with it, so the reference store's memory stops growing with deleted claims; re-saving the
+  same id does not revive the link. **Compatibility: additive.** Every read answers as before.
 
   The property exists because of what registration writes. The schema catalog goes to Drivine's
   schema manager, which ensures it on startup, so turning the flag on adds three uniqueness
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1814,20 +1814,23 @@ and the consumer PRs that deliver it).
   metadata never changes product behaviour: a run now adds a lineage write and nothing else, and a
   test runs the same extraction with and without one and compares everything except that write.
   Hosts that never passed a run get corrected edges under dedup without changing a line. Lineage is
-  written **last, after structural wiring, projection and grounding have all completed**, so that a
-  failure it raises leaves a complete extraction behind: claims persisted, structural edges wired,
-  projection run, grounding run, and no `PRODUCED_BY_RUN` edge. That is the end state a `STRICT`
-  failure reports, and `LENIENT` reaches the same one and reports success. The split of
-  `persistReturningCanonical` into `persistCanonicalPropositions` and `wireStructuralRelationships`
-  stays — both published, the original preserved as their composition — because the canonical
-  propositions a save returns are what every later pass wires against.
-  An earlier cut wrote lineage directly behind the save, ahead of the three wiring passes, so a
-  throwing projector could not leave stored claims unattributed. That ordering cannot survive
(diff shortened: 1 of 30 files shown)


jimador:
Lineage currently fails open and also changes unrelated persistence behavior:

- A missing run store returns silently, and every `RuntimeException` during attribution is swallowed. Missing or wrong-context runs can produce successful but unattributed claims.
- Merely setting `currentRun` switches persistence to the canonical saved proposition. The no-run path keeps the pre-save minted ID, so audit metadata changes projection and grounding behavior.
- Concurrent links can create duplicate `PRODUCED_BY_RUN` relationships. There is no relationship uniqueness backstop, sequential tests do not cover the race, and `DISTINCT` reads hide it.
- The run/link stores and schema are wired only by the test application, not production auto-configuration.

Canonical persistence must be unconditional. Define an explicit attribution failure policy, make link creation concurrency-safe and observable, wire the feature in production, and add a full remember-to-canonical-to-Drivine integration test.

jimador:
Stepping back from the slices to the workstream. Four things need deciding before Wave D starts.

1. Nothing in DICE produces a run. `ExtractionRunStore.save`, `recordInvocation` and `transition` have no non-test caller anywhere in the stack, and `docs/design/extraction-runs.md:1037-1041` says so. `metamodelFingerprint` has no producer either. Slices 6 through 9 are a durable store for a record that only a hand-written host can write. The coordinator (slice 10) is the first real caller of this contract and the slice that tells us whether the API shape is right. Land it before #68 and #69, and land it with a consumer path. `me` calls plain `rememberText` at four sites (`PersonaKgFeed.kt:110`, `SelfQuizRunner.kt:71`, `MemoryTools.kt:97`, `RecoveryLearningWriter.kt:58`) and nothing there mints a revision, a profile or a run. One real path, Obsidian notes with the content hash as the revision and one run per ingestion batch, exercises every contract in this stack end to end. The store contract has had many review rounds and no caller.

2. Run presence changes persistence semantics. `persistAndProject` (`IncrementalPropositionExtraction.kt:733`) switches projection and grounding onto canonical ids only when `currentRun` is set, and the no-run path keeps wiring against pre-save minted ids. The inline thread already names the defect. The rule at the architecture level is that audit metadata never changes product behaviour. Canonical persistence becomes the only path, and a run adds a lineage write and nothing else.

3. The join with the metamodel stack is undefined. `ExtractionRun.fingerprints.metamodelFingerprint` (`ExtractionRunEnvelope.kt:41`) is a pass-through string. On the other stack `dice.metamodel.version` is a proposition metadata key nothing writes (thread on #83). Both describe the same fact, which schema declaration a claim was extracted under. Decide once. The run carries the `DeclaredSchema` stamp hash, the coordinator resolves it from the `DeclaredSchemaSource` bean when one is present, and per-proposition schema attribution is answered through `PRODUCED_BY_RUN`. The metadata key then goes, or becomes a denormalised copy the coordinator writes.

4. Terminal fingerprint versus #69. `ExtractionRunTransition.fingerprint` hashes `counts` and `failures` (`ExtractionRunTransition.kt:54`), and `ExtractionRunCounts` is five plain ints. #69's typed product outcomes change that payload, which changes a persisted fingerprint format. Either hash only the transition identity (status, finishedAt) and carry counts as data, or plan on `xrun-terminal:v2` before this ships in a release. State which in the design doc.

Two smaller ones. No `DiceEvent` is emitted for any run transition, so the Wave F audit projection and any host listener have nothing to subscribe to. `ExtractionRunFingerprint.ofFields` with `HEADER_VERSION` and `INVOCATION_VERSION` has no caller because `DrivineExtractionRunStore` compares in Cypher. Drop it or use it.

Landing note. This stack and the metamodel stack conflict on `DrivinePropositionRepository.kt`, `CHANGELOG.md` and two integration tests (`git merge-tree` of the two tips). `4bf7e86` on #87 is the metamodel side, #90 to #92 are this side. Whichever lands second restacks, and the union should build once before either lands.

jimador:
All four decided; three are landed on the stack, the fourth is sequenced as you asked.

**1. The coordinator lands next, with a consumer path.** Agreed on ordering: the store contract is shipped and the coordinator is the slice that proves the API shape. It is recorded for the next wave, ahead of #68 and #69, together with the Obsidian-notes path (content hash as the revision, one run per ingestion batch) and the `me` `DeclaredSchemaSource` bean. Deliberately out of this round: it is its own slice with its own consumer, and rushing it into this stack would have shipped a second untested writer.

**2. Canonical persistence is the only path.** Landed as "Attribute claims to runs on the canonical path only". `persistAndProject` now wires projection, grounding and structural relationships against what `save` returned on every call; a run's only effect is the lineage write. Restoring the old no-run wiring fails three tests, including projection targeting a pre-save id. This is the sanctioned behavior change to the no-run path, stated in the CHANGELOG.

**3. One attribution mechanism: the run.** On #83 the metadata key is removed ("Answer schema attribution through run lineage"); the run carries the declared schema's `contentHash` and per-proposition attribution answers through `PRODUCED_BY_RUN`. `metamodelFingerprint` keeps its field, and its KDoc now names its one writer — the coordinator, resolving the host's `DeclaredSchemaSource` — so nothing else may fill it. The versioning doc says the coordinator may write a denormalised per-proposition copy once it exists.

**4. Hash transition identity only.** Landed as "Hash only transition identity and announce applied transitions". `ofTerminal` covers `(status, finishedAt)`; counts and failures travel as data on the transition and still participate in its equality. `TERMINAL_VERSION` moved to `xrun-terminal:v2` once, before any release, and the design doc states the choice — #69's typed product outcomes change no persisted fingerprint format.

**The two smaller ones.** `ExtractionRunTransitioned` is a `DiceEvent`: both stores emit it after an applied terminal transition ("Persist the failure vocabulary and announce transitions from Drivine"), gated on the result, nothing wired automatically — the audit projection and host listeners have their subscription point. `ofFields` is used: the Drivine row mappers digest header and invocation rows through it under `HEADER_VERSION` and `INVOCATION_VERSION`, backing the versioned compare-and-set header writes.

**Landing note taken.** The union of both stack tips was built green before either lands. Whichever lands second restacks and applies the one recorded adaptation — `ExtractionRunSchema` joins the metamodel stack's derived-ownership contract — and the guard test fails loudly until it is applied. `me`'s `undoSingleCollapse` call migrates per the #92 CHANGELOG.

jimador:
Second pass over the revised stack. Points 2, 3 and 4 landed as asked, the request object is the right shape, and `ExtractionRunTransitioned` from both stores with a `DEV_NULL` default is the correct opt-in. Three things to fix before this lands, all on the opt-in rule: everything in this train is experimental and a host that never asks for runs must see nothing.

1. The run stores are no longer opt-in. `DiceStorageAutoConfiguration.kt:158,174,187` now registers `DrivineExtractionRunStore`, `DrivinePropositionRunLinkStore` and an `ExtractionRunSchema` catalog under the existing `embabel.dice.store.type=graph` condition. Drivine's schema manager applies every `SchemaCatalog` bean at startup, so `me` and `assistant` get three constraints and five indexes written to their Neo4j on the next snapshot, for a feature neither calls. The metamodel stack gates its whole bean set and its DDL on the host declaring a schema. Do the same here. There is no natural bean to gate on, so use a property, `embabel.dice.extraction.runs.enabled` defaulting to false, and put the two stores and the schema catalog behind it. The CHANGELOG entry for `dice-storage-autoconfigure` then says what turning it on writes to the database.

2. The metamodel join is decided in KDoc only. `ExtractionRunEnvelope.kt` says the coordinator writes `metamodelFingerprint` from the host's `DeclaredSchemaSource`, and `docs/design/extraction-runs.md` never mentions the field. Put the decision in the design doc, in the same words as `metamodel-versioning.md:48-53` on the other stack, so the two docs describe one mechanism.

3. Experimental marking has gaps. `SourceRevisionQueryCapable`, `ProvenanceScanningSourceRevisionQueries`, `ProvenanceSubtractionCapable`, `SourceIdentityBounds`, `InvalidSourceProvenanceException` and `ExtractErrorResponse` are new public types with no `@ApiStatus.Experimental`. `SourceIdentityBounds` is the one that changes behaviour for existing callers, so it needs the marker most. While there, the `/extract` 400 body changing from empty to `{"error": ...}` applies to every rejection path on an existing endpoint and deserves its own compatibility line rather than riding the length-bound entry.

Landing note. The union now conflicts on `DiceStorageAutoConfiguration.kt` and `DrivineCollectorRecordStore.kt` as well as `DrivinePropositionRepository.kt` and the changelog. Fine as long as the recorded resolutions cover the new two.

james-gsb:
All three landed.

1. eeed0b7 puts both stores and the schema catalog behind `embabel.dice.extraction.runs.enabled`, defaulting to false, alongside the graph condition. A host that never asks sees no beans and no DDL. The `dice-storage-autoconfigure` CHANGELOG entry now lists exactly what turning it on writes: three uniqueness constraints and five range indexes, named. Gate tests assert absence by default; a mutation probe with the condition removed fails 3 of them.

2. 7d8a8dd (on #95's branch) adds "How a run joins to the metamodel" to `docs/design/extraction-runs.md`, in the same words as `metamodel-versioning.md:48-53`: the coordinator resolves the host's `DeclaredSchemaSource` and stamps `metamodelFingerprint` on the run record.

3. 9eed6a3 (#90) and f51483c (#93) mark all six types, `SourceIdentityBounds` included, and the `/extract` 400-body change now has its own compatibility line separate from the length-bound entry.

On the landing note: rerere covers `DrivineCollectorRecordStore.kt` and the changelog. Its recording for `DiceStorageAutoConfiguration.kt` turned out bad (it displaced the lineage schema beans), so it was forgotten and that conflict gets resolved by hand from the union worktree's committed adaptation, which now also carries the runs gate on both schema beans. Union of both new stack tops: full reactor test, exit 0.

jimador:
All three landed.

1. eeed0b7 puts both stores and the schema catalog behind `embabel.dice.extraction.runs.enabled`, defaulting to false, alongside the graph condition. A host that never asks sees no beans and no DDL. The `dice-storage-autoconfigure` CHANGELOG entry now lists exactly what turning it on writes: three uniqueness constraints and five range indexes, named. Gate tests assert absence by default; a mutation probe with the condition removed fails 3 of them.

2. 7d8a8dd (on #95's branch) adds "How a run joins to the metamodel" to `docs/design/extraction-runs.md`, in the same words as `metamodel-versioning.md:48-53`: the coordinator resolves the host's `DeclaredSchemaSource` and stamps `metamodelFingerprint` on the run record.

3. 9eed6a3 (#90) and f51483c (#93) mark all six types, `SourceIdentityBounds` included, and the `/extract` 400-body change now has its own compatibility line separate from the length-bound entry.

On the landing note: rerere covers `DrivineCollectorRecordStore.kt` and the changelog. Its recording for `DiceStorageAutoConfiguration.kt` turned out bad (it displaced the lineage schema beans), so it was forgotten and that conflict gets resolved by hand from the union worktree's committed adaptation, which now also carries the runs gate on both schema beans. Union of both new stack tops: full reactor test, exit 0.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/LineageFailurePolicy.kt
```diff
@@ -0,0 +1,95 @@
...
+ * warning in a log is not a control: nobody reads it in time, and the extraction that dropped the
+ * record reported success. Asking for attribution and getting none should be loud.
+ *
+ * [LENIENT] exists for the host that has decided the claims matter more than their attribution and
+ * is willing to say so in configuration. It is a deliberate downgrade, chosen once, in the open.
+ *
+ * EXPERIMENTAL. The shape may still change while extraction runs (DICE #67) land.
+ */
+@ApiStatus.Experimental
+enum class LineageFailurePolicy {
```
jimador, line 40: Lineage failed open: a missing run store returned silently, and every `RuntimeException` during attribution was swallowed. Missing or wrong-context runs could produce successful but unattributed claims. Define an explicit attribution failure policy.

jimador: Attribution fails loud. A run present with no way to record its lineage is an error by policy (`LineageFailurePolicy`), and lineage is recorded immediately after the claims are saved, before any fallible pass, so a throwing projector or grounding service leaves saved claims attributed and still propagates. The honest limit is in the shipped docs: a statement failing at the server inside a wrapping host transaction terminates that transaction beneath Spring; the structural fix belongs to the coordinator slice.

dice/src/main/kotlin/com/embabel/dice/proposition/extraction/IncrementalPropositionExtraction.kt
```diff
@@ -309,7 +402,7 @@ open class IncrementalPropositionExtraction @JvmOverloads constructor(
 
         if (result != null && result.propositions.isNotEmpty()) {
             logger.info(result.infoString(true, 1))
-            persistAndProject(result)
+            persistAndProject(result, context)
```
jimador, line 405: Merely setting `currentRun` switched persistence to the canonical saved proposition. The no-run path kept the pre-save minted id, so audit metadata changed projection and grounding behavior. The rule at the architecture level is that audit metadata never changes product behaviour. Canonical persistence must be unconditional, and a run adds a lineage write and nothing else.

jimador: **Canonical persistence is the only path.** Landed as "Attribute claims to runs on the canonical path only". `persistAndProject` now wires projection, grounding and structural relationships against what `save` returned on every call; a run's only effect is the lineage write. Restoring the old no-run wiring fails three tests, including projection targeting a pre-save id. This is the sanctioned behavior change to the no-run path, stated in the CHANGELOG.

dice-storage/src/main/kotlin/com/embabel/dice/storage/DrivinePropositionRunLinkStore.kt
```diff
@@ -0,0 +1,334 @@
...
+import org.drivine.query.QuerySpecification
+import org.slf4j.LoggerFactory
+import org.springframework.transaction.annotation.Transactional
+
+/**
+ * Drivine / Neo4j implementation of [PropositionRunLinkStore].
+ *
+ * ## The graph
+ *
+ * `(:Proposition {id, contextId})-[:PRODUCED_BY_RUN]->(:ExtractionRun {contextId, runId})`
```
jimador, line 34: Concurrent links could create duplicate `PRODUCED_BY_RUN` relationships. There was no relationship uniqueness backstop, sequential tests did not cover the race, and `DISTINCT` reads hid it. Make link creation concurrency-safe and observable.

jimador: The edge is `MERGE`-idempotent on the pair and tenant-guarded on both endpoints, with all-or-nothing batch writes proven inside the TOCTOU window; the linking-twice case is in the 14-case cross-backend contract suite. The headline invariant on Neo4j: two identical-content runs leave one proposition, one source grounding, and two run links.

dice-storage-autoconfigure/src/main/kotlin/com/embabel/dice/storage/autoconfigure/DiceStorageAutoConfiguration.kt
```diff
@@ -139,6 +144,82 @@ class DiceStorageAutoConfiguration {
         persistenceManager: PersistenceManager,
     ): CollectorRecordStore = DrivineCollectorRecordStore(persistenceManager)
 
+    /**
+     * The durable extraction-run header store.
+     *
+     * Two conditions, both required. The graph backend has to be selected
+     * (`embabel.dice.store.type=graph`), and extraction runs have to be turned on with
+     * `embabel.dice.extraction.runs.enabled=true`. The property defaults to off, so an upgrading
```
jimador, line 155: The run stores were wired only by the test application, not production auto-configuration. Wire the feature in production, and add a full remember-to-canonical-to-Drivine integration test.

Then, once wired, they were no longer opt-in: `DiceStorageAutoConfiguration` registered `DrivineExtractionRunStore`, `DrivinePropositionRunLinkStore` and an `ExtractionRunSchema` catalog under the existing `embabel.dice.store.type=graph` condition. Drivine's schema manager applies every `SchemaCatalog` bean at startup, so every graph-backed host gets three constraints and five indexes written to its Neo4j on the next snapshot, for a feature it never calls. The metamodel stack gates its whole bean set and its DDL on the host declaring a schema. Do the same here. There is no natural bean to gate on, so use a property, `embabel.dice.extraction.runs.enabled` defaulting to false, and put the two stores and the schema catalog behind it. The CHANGELOG entry for `dice-storage-autoconfigure` then says what turning it on writes to the database.

jimador: eeed0b7 puts both stores and the schema catalog behind `embabel.dice.extraction.runs.enabled`, defaulting to false, alongside the graph condition. A host that never asks sees no beans and no DDL. The `dice-storage-autoconfigure` CHANGELOG entry now lists exactly what turning it on writes: three uniqueness constraints and five range indexes, named. Gate tests assert absence by default; a mutation probe with the condition removed fails 3 of them. The remember-to-canonical-to-Drivine path is the Neo4j integration test behind the headline invariant.

docs/design/extraction-runs.md
```diff
@@ -6,10 +6,11 @@ model for, what the provider actually reported back, how far it got, and what we
...
 objects, no provider SDK payloads, no extension maps.
 
-This note covers DICE #67's value model and its store contract — the types in
-`com.embabel.dice.proposition.extraction`, the lifecycle state machine that governs a run's status,
-and the reads a store owes. The Drivine implementation, the proposition-to-run relation and the
-wiring are separate slices.
+This note covers DICE #67's value model, its store contract, the Drivine implementation, and the
+proposition-to-run relation — the types in `com.embabel.dice.proposition.extraction`, the lifecycle
+state machine that governs a run's status, the reads a store owes, and how a stored claim gets
+traced back to the runs that produced it. The run coordinator and its REST exposure are a separate
```
jimador, line 18: Nothing in DICE produces a run. `ExtractionRunStore.save`, `recordInvocation` and `transition` have no non-test caller anywhere in the stack, and this doc says so. `metamodelFingerprint` has no producer either. Slices 6 through 9 are a durable store for a record that only a hand-written host can write. The coordinator (slice 10) is the first real caller of this contract and the slice that tells us whether the API shape is right. Land it before #68 and #69, and land it with a consumer path. The known host calls plain `rememberText` at four sites and nothing there mints a revision, a profile or a run. One real path, Obsidian notes with the content hash as the revision and one run per ingestion batch, exercises every contract in this stack end to end. The store contract has had many review rounds and no caller.

jimador: **The coordinator lands next, with a consumer path.** Agreed on ordering: the store contract is shipped and the coordinator is the slice that proves the API shape. It is recorded for the next wave, ahead of #68 and #69, together with the Obsidian-notes path (content hash as the revision, one run per ingestion batch) and the host's `DeclaredSchemaSource` bean. Deliberately out of this round: it is its own slice with its own consumer, and rushing it into this stack would have shipped a second untested writer.

CHANGELOG.md
```diff
@@ -995,3 +996,181 @@ and the consumer PRs that deliver it).
   Reads are uncapped, because a read for a longer tenant matches nothing by construction. Every new
   type carries `@ApiStatus.Experimental` and the shapes may still move while the remaining #67 slices
```
jimador, line 1715: Experimental marking had gaps. `SourceRevisionQueryCapable`, `ProvenanceScanningSourceRevisionQueries`, `ProvenanceSubtractionCapable`, `SourceIdentityBounds`, `InvalidSourceProvenanceException` and `ExtractErrorResponse` were new public types with no `@ApiStatus.Experimental`. `SourceIdentityBounds` is the one that changes behaviour for existing callers, so it needs the marker most. While there, the `/extract` 400 body changing from empty to `{"error": ...}` applies to every rejection path on an existing endpoint and deserves its own compatibility line, separate from the length-bound entry.

jimador: 9eed6a3 (#90) and f51483c (#93) mark all six types, `SourceIdentityBounds` included, and the `/extract` 400-body change now has its own compatibility line separate from the length-bound entry.

CHANGELOG.md
```diff
@@ -995,3 +996,181 @@ and the consumer PRs that deliver it).
...
+  one batch names an id twice — two revision results touching one original — every position reports
+  the store's *last* answer for it, because a replace-by-id store overwrites the first and the first
+  object is stale from that moment; resolving rather than rejecting matters because `of` runs after
+  the saves, so throwing would fail an extraction whose propositions are already written. One input
+  id answered with two *different* stored ids is still rejected.
+  `DrivinePropositionRepository` itself needed no change — the canonical id was already in its
+  return value, and it was being dropped a layer up. `EventEmittingPropositionRepository` overrides
+  the new call for the same reason it overrides `saveAll`: Kotlin's `by delegate` forwards an
+  interface default straight past the decorator's own `save`.
+  **The relation.** `(:Proposition {id, contextId})-[:PRODUCED_BY_RUN]->(:ExtractionRun {contextId,
```
jimador, line 1757: Landing note. This stack and the metamodel stack conflict on `DrivinePropositionRepository.kt`, `CHANGELOG.md` and two integration tests (`git merge-tree` of the two tips), and later also on `DiceStorageAutoConfiguration.kt` and `DrivineCollectorRecordStore.kt`. Whichever lands second restacks, and the union should build once before either lands, with the recorded resolutions covering all of them.

jimador: **Landing note taken.** The union of both stack tips was built green before either lands: full reactor test, exit 0. Whichever lands second restacks and applies the one recorded adaptation, `ExtractionRunSchema` joining the metamodel stack's derived-ownership contract, and the guard test fails loudly until it is applied. rerere covers `DrivineCollectorRecordStore.kt` and the changelog. Its recording for `DiceStorageAutoConfiguration.kt` turned out bad (it displaced the lineage schema beans), so it was forgotten and that conflict is resolved by hand from the union worktree's committed adaptation, which also carries the runs gate on both schema beans. A downstream caller of `undoSingleCollapse` migrates per the #92 CHANGELOG, at its convenience since the four-argument form stays as a deprecated overload.

igordayen:
Findings from CODEX:

Findings

  1. STRICT lineage does not fail the async event path.
     IncrementalPropositionExtraction.kt:472-476 (https://github.com/embabel/dice/pull/101/files#diff-1) catches every exception from
     persistAndProject, including LineageNotRecordedException, and only logs it. But SourceAnalysisRequestEvent.kt:67-77
     (https://github.com/embabel/dice/pull/101/files#diff-2) now lets async publishers attach currentRun(). Example: an event carries
     currentRun, no PropositionRunLinkStore is bound, STRICT throws, then the listener swallows it. The caller/coordinator sees no
     failure signal, so “strict attribution fails the operation” is only true for direct calls, not for the async path added here.

  2. The current ordering can leave saved claims unattributed when projection/grounding fails.
     IncrementalPropositionExtraction.kt:611-647 (https://github.com/embabel/dice/pull/101/files#diff-3) saves canonical propositions,
     then structural wiring, projection, grounding, and only then writes lineage. So if projection throws after the save, the
     proposition is stored but no PRODUCED_BY_RUN edge is attempted. Example: run R1 extracts proposition P1; save succeeds; projector
     throws; P1 remains in the store, but propositionsOf(R1) is empty. This contradicts the PR body and
     PersistablePropositions.kt:120-125 (https://github.com/embabel/dice/pull/101/files#diff-4), which say the split exists so saved
     claims are not stranded without the run that produced them. If lineage-last is intentional, those statements need to be corrected;
     if traceability is the invariant, lineage needs to happen immediately after canonical save.

  3. InMemoryPropositionRunLinkStore retains stale links forever.
     InMemoryPropositionRunLinkStore.kt:57-58 (https://github.com/embabel/dice/pull/101/files#diff-5) keeps every run-to-proposition id
     in byRun, and lines 108-110 (https://github.com/embabel/dice/pull/101/files#diff-5) explicitly leave deleted proposition ids in
     the map. Reads filter stale ids, but memory still grows with every extraction/run/proposition ever linked. For a long-running
     in-memory host or tests that reuse the store, deleted or expired propositions disappear from API results but remain retained
     internally. Consider opportunistic pruning on reads, a clear/delete hook, or making this explicitly test/dev-only.

igordayen (commented):
@jimador - would you please review some findings from Codex, concern with stale memory, thank you

jimador:
All three addressed.

1. Lineage now runs right after the canonical save, before structural wiring, projection and grounding, so a claim is attributed the moment it exists and a throwing projector cannot strand it. The KDoc and the design doc describe the state a STRICT failure leaves and no longer argue for lineage-last. d3cbee7e.
2. The async path honours STRICT: a `LineageNotRecordedException` is logged at error and rethrown to the publisher, and the queue keeps draining so one failing event does not drop the next. f8eb4f7d.
3. The in-memory link store prunes an id on read when the proposition no longer exists at all, and drops an empty run entry; a read from the wrong tenant still fails closed and touches nothing. af3ff709.

igordayen:
> 3\. The in-memory link store prunes an id on read when the proposition no longer exists at all, and drops an empty run entry; a read from the wrong tenant still fails closed and touches nothing. [af3ff7](https://github.com/embabel/dice/commit/af3ff70939dd2eed72cb82e7e458e91fad594098)

@jimador - would an eviction policy be useful for this use case? Thank you

igordayen (approved):
@jimador - thanks for addressing inquiries, looks good. thanks

igordayen:
@jimador - looks like rebase is required, thx

jimador:
Yes, and it is in ff8e3438. Pruning on read only helps a run somebody reads, so `InMemoryPropositionRunLinkStore` now takes `maxRuns` (default 10,000, same as the reference run store) and past it drops the links of the runs it linked earliest once the run store says they have ended. A run still `RUNNING` keeps its links, and a breach with nothing to evict logs once. Design doc and changelog updated.

On the rebase: the branch is on current main, GitHub reports it clean. What it is showing is the stacked base, since #101 sits on #99, which sits on #98 and so on down to #91. It merges after those, and the stack tool moves each remaining branch onto main as the one below it lands.

PR #102, opened 2026-09-02
REST controller conditions answered after auto-configuration

jimador:
`@ConditionalOnBean` on a plainly imported class is answered while the importing configuration class is being read, before Spring Boot registers any auto-configuration bean definition. So `DiscoveryController` and `PropositionPipelineController`, both named in `DiceRestConfiguration`'s `@Import` list with conditions on store beans, silently dropped their routes in the documented wiring: stores from `dice-storage-autoconfigure`, runner and pipeline hand-built. The repro showed the pipeline controller affected too — its host-supplied gate bean sat in a sibling import Spring processed after the condition was asked, so the plain-import shape is order-fragile even against host beans.

The fix: `MemoryController` (unconditional) stays a direct import; the two conditional controllers arrive through `ConditionalControllerImport`, an internal `DeferredImportSelector` at lowest precedence, behind Boot's own auto-configuration selector. By the time each condition is asked, its beans are on the registry or never will be. No new public API, no properties; the host-facing idiom is still one `@Import(DiceRestConfiguration.class)`.

Evidence: the two discriminating tests fail against unmodified main (routes read off a live `RequestMappingHandlerMapping`, so failure means a real 404) and pass with the fix. Five wiring tests total: both route tests, memory riding the same import, the no-import negative, and a missing-runner negative proving a clean context with discovery absent.

Same measurement and idiom as the governance activation work on #88; when that lands, `GovernanceController` folds into this selector's array as one line.

### DIFF
--- a/dice-storage-autoconfigure/pom.xml
+++ b/dice-storage-autoconfigure/pom.xml
@@ -164,6 +164,30 @@
             <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <!--
+          Test: WebApplicationContextRunner needs a servlet API to build its mock context, and that
+          is how the REST controllers' web conditions are checked. Runtime needs nothing new: the
+          controllers live in dice, which already declares the servlet API itself.
+        -->
+        <dependency>
+            <groupId>jakarta.servlet</groupId>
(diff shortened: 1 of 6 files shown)


igordayen (approved):
@jimador - looks good, thanks

PR #103, opened 2026-09-02
Record collector decisions from the duplicate strategy

jimador:
## The bug

A sweep run with `DuplicateCollectorStrategy` wrote `CollectorRecord` rows naming the survivor and no `CollectorDecision`. `undoSingleCollapse` reads the collapse through `CollectorTraceQuery.findDecisionForProposition`, so a cosine collapse was applied and could not be undone. Only `MultiSignalCollectorStrategy` recorded decisions.

Observed live on 2026-09-02: `strategyName: "duplicate"`, trace endpoint `{"decisions":[],"edges":[]}`, unmerge 404 on the first call.

## The change

- `DuplicateCollectorStrategy` takes an optional third constructor argument, `traceStore: CollectorTraceStore?`, under `@JvmOverloads`, and implements `RunAwareCollectorStrategy` so it sees the run id. On a run with a non-blank run id it records the run context, one candidate edge per clustered pair (the repository's cosine score as the single signal), the union-find components, and one `CollectorDecision` per collapsed component with a `RetiredProposition` per loser. A blank run id (the read-only `collect()` path) records nothing. Trace writes are best effort, logged on failure.
- `retiredByFold` (new, `internal`, `collector/FoldedEvidence.kt`) computes what a loser adds to the survivor: exclusive grounding, provenance refs, evidence keys, source ids, prior status. `MultiSignalCollectorStrategy` now calls it too, so the two strategies share one rule.
- `CHANGELOG.md` under Unreleased / Fixed; one sentence in `docs/design/reclamation-and-collector.md`.

**Compatibility: additive.** The one- and two-argument constructors are unchanged. A strategy built without a trace store behaves as before.

## Tests

| Suite | Result |
| --- | --- |
| `CollectorUndoCapabilityTest` (+2: a duplicate-strategy collapse run through `DefaultCollectorRunner` and undone; the untraced sweep leaves nothing to undo) | 32/32 |
| `DuplicateCollectorStrategyTest` (+2: decision contents; blank run id records nothing) | 10/10 |
| `dice` module | 1535 tests, 0 failures |

Live, against Neo4j through a host application on the reinstalled jar: a real sweep collapsed a paraphrase; the trace endpoint returned one decision and one cosine edge; unmerge answered 200 and the member went STALE to ACTIVE with the survivor's evidence untouched and `undoneAt` stamped; a second unmerge answered 404; search listed the restored memory.

## Host follow-up

Dice's autoconfiguration only builds the multi-signal strategy. A host that builds `DuplicateCollectorStrategy` itself passes its `CollectorTraceStore` bean when constructing it.

## Review round

Adversarial review found no blocking issues. One medium finding fixed: when a cluster's anchor sits outside the candidate snapshot, the union-find still groups that cluster's candidate members, and no edge was recorded to explain the component. Edges are now emitted for every clustered pair that reached a candidate, with a unit test for that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -728,6 +728,23 @@ and the consumer PRs that deliver it).
   moved to `QUARANTINED`. **Compatibility: additive.** No shipped symbol changes; the module now
   builds from a clean tree.
 
+- `DuplicateCollectorStrategy` records a `CollectorDecision` for every component it collapses, given
+  a `CollectorTraceStore`. It takes the store as a third, optional constructor argument, the same
+  way `MultiSignalCollectorStrategy` does, and implements `RunAwareCollectorStrategy` so it sees the
+  run id. Before this, a sweep run with the duplicate strategy wrote `CollectorRecord` rows naming
+  the survivor and no decision at all, so `undoSingleCollapse` and anything reading
+  `CollectorTraceQuery.findDecisionForProposition` found nothing to reverse: the collapse was
+  applied and could not be undone. Each decision carries one `RetiredProposition` per loser with the
+  grounding, provenance refs, evidence keys, source ids and prior status the merging sweep folds
+  onto the survivor, computed by the same rule the multi-signal strategy uses (only what the
+  survivor did not already hold). The cluster similarities are recorded as candidate edges and the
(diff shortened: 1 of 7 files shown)


igordayen (approved):
@jimador - looks good, thanks

PR #113, opened 2026-09-20
The proposition vector index can be made after startup, when a model appears

### LINKED ISSUE #112: The proposition vector index waits for a boot that no longer happens
## What happens

`DiceStorageAutoConfiguration.propositionVectorIndexSchema` registers an empty catalog when there is no embedding model, with this reasoning:

> A vector index is created AT the embedding model's dimension, so with no model there is no dimension to create it at. Register nothing rather than guess: an index at the wrong dimension is worse than none, because writes to it succeed. **The catalog is rebuilt on the next boot, by which time a model configured at first run is registered.**

That last sentence was true while first-run setup ended in a process restart. The Embabel appliance stopped restarting (embabel/me#1410 — connecting a provider now takes effect per call), so **"the next boot" never comes**.

## Observed

A fresh appliance, a provider key given during setup, then a document uploaded:

    06:14:43  Skipping proposition vector index schema: no embedding model
    06:15:08  Key accepted by 'openai'
    07:55:19  Embedding with 'text-embedding-3-small' from the stored 'openai' key
              Failed to invoke procedure db.index.vector.queryNodes:
              There is no such vector schema index: Proposition_embedding_vector

The embedding model **does** become available after startup and is used. The index is simply never created, so retrieval fails for the life of the process.

`ChatStoreAutoConfiguration` skips its chat-message vector index on the same condition and presumably has the same gap.

## What is needed

A way to converge this index **after** startup, once a model exists — the host knows when that is (the appliance publishes an event when a provider connects and already has a listener that provisions the RAG store).

Drivine has the right primitive: `SchemaManager.enforce()` is documented as idempotent and version-aware. The obstacle is that catalogs are fixed at construction, so re-enforcing applies the empty catalog this bean registered.

Two shapes, and the choice is yours rather than mine:

1. **Dice exposes an ensurer** — a small bean with `ensure(): Boolean` that resolves the embedding service, builds `propositionVectorIndexSpec(dimensions)` and applies it idempotently. The host calls it when a model appears. Needs an injectable index/schema manager: `IndexManager`'s constructors are `internal`, so dice cannot build one today.
2. **The catalog becomes deferred** — the spec is produced when the schema is applied rather than when the bean is built, so an existing `enforce()` after a model arrives does the right thing with no new API.

## Why I did not send a patch

Both shapes need a Drivine seam I could not see from the `drivine4j` checkout: `SchemaManager` is constructed only in tests there, so its production wiring lives in a starter outside it. Guessing at that wiring seemed worse than describing the problem exactly.

Happy to implement either once the seam is settled.

## Meanwhile

embabel/me#1436 fixes the neighbouring half — the appliance was not announcing a provider connected during setup at all, so even the RAG store was never provisioned. Documents work after it; **memory still needs a restart** until this is addressed.

jasperblues:
Closes #112.

## The assumption that stopped holding

`propositionVectorIndexSchema` registers nothing when there is no embedding model, and **that part is right** — the index is created at the model's dimension, and one made at a guessed dimension accepts writes a real model would disagree with. What stopped being true is the sentence after it:

> The catalog is rebuilt on the next boot, by which time a model configured at first run is registered.

That held while a host without a model restarted once one was configured. The Embabel appliance stopped restarting — a provider key now takes effect per call (embabel/me#1410) — so **the next boot never comes**.

Observed on a fresh appliance, key given during setup, document uploaded:

    Skipping proposition vector index schema: no embedding model
    Key accepted by 'openai'
    Embedding with 'text-embedding-3-small' from the stored 'openai' key
    Failed to invoke db.index.vector.queryNodes:
    There is no such vector schema index: Proposition_embedding_vector

The model comes alive and is used. The index is simply never made.

## Why not a catalog

`SchemaCatalog` holds **materialised** specs — no supplier or deferred form — and `SchemaManager` snapshots the catalog beans when it is constructed (`catalogs.orderedStream().toList()` in `DrivineSchemaConfiguration`). So a later `enforce()` applies the same empty catalog, and a catalog registered afterwards is never seen.

This goes through `PersistenceManager.indexes` instead, which is the public way to converge a single index with no catalog at all. **It needs nothing new from Drivine.**

## Who calls it

The host, because only the host knows a model has appeared — it is the one that took the key. Dice offers the capability and no trigger: no polling, no listener for an event this library cannot see, nothing that runs unasked.

`IndexManager.ensure` is idempotent, so a host may call it on every provider change without checking first. `false` means there is still no model — the ordinary answer on a host without a key, not an error.

## Tests

Five, including the two that would otherwise be assumed rather than checked:

- the dimension is read **per call**, so a model arriving after the bean was built is seen — this is the whole point of the class
- a **placeholder** service that throws rather than inventing a dimension counts as no model, exactly as absence does
- a failure to converge is reported and **not** thrown at the host, which is calling from wherever it noticed a key arrive

## Host side

embabel/me will call this from the listener it already has for provider-connected (`KeyedEmbeddingProvisioning`). I will raise that once this is released — happy to prepare it in parallel if you would rather land them together.


jimador (approved):
Reviewed and ran the module locally: all 5 tests in PropositionVectorIndexConvergenceTest pass, and the reasoning about SchemaManager snapshotting the catalog beans matches what's in DrivineSchemaConfiguration. Going through PersistenceManager.indexes is the right seam.

One nit, not blocking: the new bean's KDoc landed between the existing KDoc for propositionVectorIndexSpec and that function, so the spec's doc block is now orphaned above the convergence bean. Worth a follow-up shuffle.

